Issue #3, opened 2019-11-20
AWS::CloudFormation::Type


pgarbe:
## 1. Title 
AWS::CloudFormation::Type

## 2. Scope of request 
AWS::CloudFormation::Type - can create resource via API, but not via CloudFormation

## 3. Expected behavior
In Create, the RegisterType should be registered, and in Delete it should be unregistered. It seems that updates are not supported.

## 4. Suggest specific test cases
Just being able to do the same stuff as with the API / CLI.

## 5. Helpful Links to speed up research and evaluation
See `registerType` and `deregisterType` in https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/CloudFormation.html

## 6. Category (required) - Will help with tagging and be easier to find by other users to +1

5. Management (CloudTrail, Config...)

rjlohan:
> * Why is `Kind` a property? Why not have the resource type be `AWS::CloudFormation::Resource`? I get that the API is RegisterType with a `Type` field in it, but I think it's better to reify them into proper CloudFormation types than have this weird subtyping going on.

Right now the Registry only supports the ability to (De)RegisterType, however we designed it intentionally to be as agnostic of CloudFormation (the service) as we could. Certainly there will be abstraction leaks, but we intend to separate concerns so that we have the ability to cleanly build out other capabilities in the future. I'm OK with modelling the current `RESOURCE` kind as a discrete entity, though the intent is for the registration process to not really care that the end-user is modelling a CloudFormation Resource or some other entity.


> * You're falling into the same trap that Lambda has [where you can't create two versions in a single stack](https://read.acloud.guru/cloudformation-is-an-infrastructure-graph-management-service-and-needs-to-act-more-like-it-fa234e567c82)

For now, we have a rolling, immutable versioned history of the evolution of a `Type`. I don't believe this to be a one-way door so that we can introduce semantic versioning/labels later. We're not there yet.

> * You should require `DeletionPolicy: Retain` on these resources, since they create versions but do not delete them.

That _would_ be a leaky abstraction. The Resource platform does not know anything about CloudFormation's stack-level concerns. (To be fair a couple of things leaked through, but again the intent was to design the Registry and the resource platform to be completely oblivious to CloudFormation the service.

> * I agree with @pgarbe that it's strange to have the `DefaultVersion` on the resource representing a different version from the version that the resource itself represents. I think there are two non-exclusive options:
>   
[...]

PR #4, opened 2019-11-20
Resource implementations for AWS::CloudFormation::ResourceVersion and AWS::CloudFormation::ResourceDefaultVersion

rjlohan:
*Issue #, if available:* #3 

*Description of changes:*

Schema and handler implementations for `AWS::CloudFormation::ResourceVersion` and `AWS::CloudFormation::ResourceDefaultVersion`.

Sample template for use of these type;

```
Resources:
    InitialType:
        Type: AWS::CloudFormation::ResourceVersion
        Properties:
            TypeName: Sample::CloudFormation::Resource
            SchemaHandlerPackage: s3://cloudformationmanageduploadinfrast-artifactbucket-123456789012abcdef/sample-cloudformation-resource.zip
    UpdatedType:
        Type: AWS::CloudFormation::ResourceVersion
        Properties:
            TypeName: Sample::CloudFormation::Resource
            SchemaHandlerPackage: s3://cloudformationmanageduploadinfrast-artifactbucket-123456789012abcdef/sample-cloudformation-resource-update.zip
        DependsOn: InitialType

    DefaultVersion:
        Type: AWS::CloudFormation::ResourceDefaultVersion
        Properties:
            Arn: !Ref UpdatedType
```

Pending contract test verification.


aws-cloudformation-type/aws-cloudformation-type.json
```diff
@@ -0,0 +1,182 @@
...
+        "SourceUrl": {
+            "description": "The URL of the source code for the type.",
+            "maxLength": 4096,
+            "type": "string"
+        },
+        "TimeCreated": {
+            "description": "When the specified type version was registered.",
+            "type": "string"
+        },
+        "Type": {
```
benkehoe: Why can't this be moved from a property of the resource to be the actual resource type? `AWS::CloudFormation::Resource`?

aws-cloudformation-type/aws-cloudformation-type.json
```diff
@@ -0,0 +1,182 @@
...
+            "additionalProperties": false
+        }
+    },
+    "properties": {
+        "Arn": {
+            "description": "The Amazon Resource Name (ARN) of the type.",
+            "pattern": "^arn:aws[A-Za-z0-9-]{0,64}:cloudformation:[A-Za-z0-9-]{1,64}:([0-9]{12})?:type/.+$",
+            "type": "string"
+        },
+        "DefaultVersionId": {
```
benkehoe: It's strange to have the DefaultVersion on the resource representing a different version from the version that the resource itself represents. I think there are two non-exclusive options:
- There's a property on the resource that enables the "update default version" behavior
- There's a separate resource type for aliases, of which there's only the "default" today.

aws-cloudformation-type/aws-cloudformation-type.json
```diff
@@ -0,0 +1,182 @@
...
+        }
+    },
+    "required": [
+        "SchemaHandlerPackage",
+        "Type",
+        "TypeName"
+    ],
+    "readOnlyProperties": [
+        "/properties/Arn",
+        "/properties/TypeVersionArn",
```
benkehoe: Shouldn't there also be a `VersionId` so that it could be copied for use with `DefaultVersionId`?

aws-cloudformation-type/aws-cloudformation-type.json
```diff
@@ -0,0 +1,182 @@
...
+            ],
+            "type": "string"
+        },
+        "Schema": {
+            "description": "The schema that defines the type.\n\nFor more information on type schemas, see Resource Provider Schema in the CloudFormation CLI User Guide.",
+            "maxLength": 16777216,
+            "minLength": 1,
+            "type": "string"
+        },
+        "SchemaHandlerPackage": {
```
benkehoe: Should this have a more specific schema given that it is supposed to be an S3 bucket URL? Can it reference the S3 Bucket resource schema?

aws-cloudformation-type/aws-cloudformation-type.json
rjlohan: If we had this property on the `Type` resource itself I think it would be a `readOnlyProperties` annotation for use in Drift Detection type scenarios.

aws-cloudformation-type/aws-cloudformation-type.json
rjlohan: Yes that's the eventual intent, we just have to finish up the schema hosting work so that other schemas are accessible via URI. Not far off, and then we can update the existing schemas with the correct `$ref` pointer for these sort of properties.

aws-cloudformation-type/aws-cloudformation-type.json
rjlohan: Possibly, yes. Think we need to work out if this resource should model the `Type` or the `TypeVersion`, I think the way it will be used will be slightly different depending on which way we go.

aws-cloudformation-type/aws-cloudformation-type.json
rjlohan: Definitely could, but at this point I don't think there's anything about this model which is resource specific and I don't foresee anything like that coming in. But that sort of thinking led to AWS::RDS::DBInstance which I really want to refactor.

pgarbe:
What if you rename `Type` into `TypeVersion` and introduce another resource called `Type` which just refers to an existing `TypeVersion`?

So you can have multiple `TypeVersion`s to register different versions and only one `Type` for a specific TypeName. 

Although, I'm not really aware of an use case that needs multiple versions registered. 

```yaml
MyType:
  Type: AWS::CloudFormation::Type
  Properties:
    Version: !Ref MyTypeVersion1
    TypeName: Organisation::Service::Resource
    ExecutionRoleArn: arn:aws:iam::123456789012:role/MyExecutionRole
    LoggingConfiguration:
        LogRoleArn: arn:aws:iam::123456789012:role/MyLoggingRole
        LogGroupName: MyResourceLogGroups  

MyTypeVersion1:
  Type: AWS::CloudFormation::TypeVersion
  Properties:
    Kind: Resource
    TypeName: Organisation::Service::Resource
    DefaultVersion: "00000001"
    SchemaHandlerPackage: my-artifact-bucket

```

benkehoe:
In my ideal world, it's three resource types, `AWS::CloudFormation::Resource`, `AWS::CloudFormation::ResourceVersion`, and `AWS::CloudFormation::ResourceDefaultVersion`/`AWS::CloudFormation::ResourceVersionAlias`. The template would look like:

```yaml
MyResource:
  Type: AWS::CloudFormation::Resource
  Properties:
    TypeName: Organisation::Service::Resource

MyResourceVersion:
  Type: AWS::CloudFormation::ResourceVersion
  Properties:
    ResourceType: !Ref MyResource
    ExecutionRoleArn: arn:aws:iam::123456789012:role/MyExecutionRole
    LoggingConfiguration:
      LogRoleArn: arn:aws:iam::123456789012:role/MyLoggingRole
      LogGroupName: MyResourceLogGroups  
    SchemaHandlerPackage: my-artifact-bucket

MyResourceDefaultVersion:
  Type: AWS::CloudFormation::ResourceDefaultVersion
  Properties:
    ResourceType: !Ref MyResource
    DefaultVersion: !GetAtt MyResourceVersion.VersionId
```

This would allow multiple versions to be defined in a single template. This may become important when versioning gains more semantics (aliases representing major versions). I take the view that if you can't represent any state of the service achievable through the console in a single template, the resource model is broken. That said, it's verbose, and maybe there's a way to define `AWS::CloudFormation::Resource` such that it could take the version config inline, and later be extended to support the placeholder usage with a separate `ResourceVersion` that I've outlined above. I do think the default version should exist as a separate resource from the start, though.

rjlohan:
Thanks for the feedback @pgarbe and @benkehoe. Let's experiment with these ideas. I'm going to be somewhat offline for a couple of days, but I'll aim to try and bootstrap the repo with API calls and setup so we can see how this plays out. Or if one of you want to have a crack at a PR, you could refer to one of the working examples like [AWS::Logs::LogGroup](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-logs/tree/master/aws-logs-loggroup) for some of the conventions we use as a starting point.

aws-cloudformation-type/aws-cloudformation-type.json
aygold92: this is also a ReadOnly property -- we want to provide all the properties in the desribe-type API output

aws-cloudformation-type/aws-cloudformation-type.json
aygold92: The way the schema was written, it was meant to model `Type`

rjlohan:
> This falls into the same trap that Lambda has [where you can't create two versions in a single stack](https://read.acloud.guru/cloudformation-is-an-infrastructure-graph-management-service-and-needs-to-act-more-like-it-fa234e567c82).

This is actually not possible in the underlying APIs anyway; one registration is allowed to be in flight at the same time, and (currently) only one version can be set as the `Default` (i.e; the one which CloudFormation will invoke when this type is specified in a template)

benkehoe:
> This is actually not possible in the underlying APIs anyway
Sure. And you'll argue that since you can't deregister a type unless it has only one version, even though a type with lots of versions would require a long sequence of stack updates to restore, deregistering it by accident is unlikely and so we don't need to account for that case. And so given the API I guess it's probably ok to leave it without multiple version representation. But I do want to register my disappointment that _even on the CloudFormation team_ you're not starting with CloudFormation resource design, ensuring proper infrastructure as code lifecycle management, and working backwards from there to API design.

The fact that there can be only one default would point to `DefaultVersion` being set on the `AWS::CloudFormationResource` resource itself, but it still feels weird to have it potentially point to a different version that is represented by the resource in the stack (i.e., the current version). So I guess I think I would propose the either/or scenario I mentioned above:
* An optional `SetDefaultToLatest: True` property on the `AWS::CloudFormation::Resource`. When this is set the version represented by the resource is also the default, so there isn't that disconnect
* A separate `AWS::CloudFormation::ResourceDefaultVersion` or `AWS::CloudFormation::ResourceVersionAlias` resource that can either `!GetAtt` the current resource version or set it to a specific value.
It would be up to the user to not do both at the same time. CloudFormation linting could help with this.

aws-cloudformation-type/aws-cloudformation-type.json
benkehoe: It wasn't indicated as ReadOnly in the example @rjlohan provided in [#3](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-cloudformation/issues/3#issuecomment-555628238), it was suggested that you'd be able to set the default version through this same resource. 

Also, note that while I don't think having both API calls in the same resource is a good idea in this case ([see below](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-cloudformation/pull/4#issuecomment-558875313)), I specifically advocate discarding the notion that a resource should correspond exactly to a single API call.

aws-cloudformation-type/aws-cloudformation-type.json
benkehoe: Except that with every update, it creates a new `TypeVersion`, which we need to be able to discover and reference, for example to set the default version.

pgarbe:
> This is actually not possible in the underlying APIs anyway; one registration is allowed to be in flight at the same time, and (currently) only one version can be set as the `Default` (i.e; the one which CloudFormation will invoke when this type is specified in a template)

That's even more confusing. From the CLI, I'd assume that you can register multiple versions for a type and point the "default" to one of the registered versions. Can you elaborate what you mean by this is not possible?

benkehoe:
> From the CLI, I'd assume that you can register multiple versions for a type and point the "default" to one of the registered versions.

That's true. What Ryan was saying is that you can't have two simultaneous RegisterType calls trying to create new versions at the same time, which is what might happen if there is a resource for `AWS::CloudFormation::ResourceVersion` and you've got two of those in the same template, which would be possible under [my proposed design](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-cloudformation/pull/4#issuecomment-557097549) on purpose (I want this kind of functionality for Lambda resources).

rjlohan:
FYI I've added a basic implementation against this schema, haven't tested it yet. Just wanting to show progress and that I haven't walked away. :)

aws-cloudformation-type/src/main/java/software/amazon/cloudformation/type/CreateHandler.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    private void registerType(final ResourceModel model,
+                              final CallbackContext callbackContext) {
+        try {
+            proxy.injectCredentialsAndInvokeV2(Translator.translateToCreateRequest(model),
+                ClientBuilder.getClient()::registerType);
+        } catch (final CfnRegistryException e) {
+            throw new CfnGeneralServiceException(e);
+        }
```
aygold92: we could also catch ValidationException and return InvalidRequest

aws-cloudformation-type/src/main/java/software/amazon/cloudformation/type/DeleteHandler.java
```diff
@@ -0,0 +1,47 @@
...
+
+        try {
+            proxy.injectCredentialsAndInvokeV2(Translator.translateToDeleteRequest(model),
+                ClientBuilder.getClient()::deregisterType);
+            logger.log(String.format("%s [%s] successfully deleted.",
+                ResourceModel.TYPE_NAME, model.getPrimaryIdentifier()));
+        } catch (final TypeNotFoundException e) {
+            throwNotFoundException(model);
+        } catch (final CfnRegistryException e) {
+            throw new CfnGeneralServiceException("DeregisterType: " + e.getMessage());
```
aygold92: nit: something a little more customer friendly like `"Failed to deregister type: %s"`

aws-cloudformation-type/src/main/java/software/amazon/cloudformation/type/CreateHandler.java
```diff
@@ -0,0 +1,103 @@
...
+                    throw new CfnGeneralServiceException(response.progressStatusAsString());
+                }
+            } catch (CfnNotFoundException e) {
+                logger.log(request.getDesiredResourceState().getPrimaryIdentifier() +
+                    " does not exist; retrying stabilization.");
+                return ProgressEvent.<ResourceModel, CallbackContext>builder()
+                    .callbackContext(context)
+                    .callbackDelaySeconds(5)
+                    .status(OperationStatus.IN_PROGRESS)
+                    .resourceModel(request.getDesiredResourceState())
```
aygold92: nice. Because the Type/TypeName is the physicalId, this will satisfy the physicalId requirement, and because we don't return until we've successfully registered, we know that there is in fact a registration in progress.

One scenario that could cause leaking is if the describe call fails for some reason, but we can iterate on that as it comes up

aws-cloudformation-type/src/main/java/software/amazon/cloudformation/type/Translator.java
```diff
@@ -0,0 +1,108 @@
...
+        if (model.getLoggingConfig() != null) {
+            builder.loggingConfig(translateToSDK(model.getLoggingConfig()));
+        }
+
+         return builder.build();
+    }
+
+    static DescribeTypeRequest translateToReadRequest(@NonNull final ResourceModel model) {
+        return DescribeTypeRequest.builder()
+            .typeName(model.getTypeName())
```
aygold92: not totally sure, but I think since you're only specifying the typeName instead of the typeArn, `Type RESOURCE` is required

aws-cloudformation-type/src/main/java/software/amazon/cloudformation/type/CreateHandler.java
aygold92: Other easy one would be to check for status code > 500 and return `ServiceInternalError`

we probably also want to set a good example here and write some mapping for all the rest of the error codes.  `AccessDenied`, `InvalidCredentials`, `Throttling`, etc.

rjlohan:
So the PR currently 'works' here, except that the `AWS::CloudFormation::TypeVersion` is unable to get the `VersionId` from a `!GetAtt` to an `AWS::CloudFormation::Type`, due to a problem with the `DescribeType` API which I am going to fix in the Registry. You should be able to hardcode a version ID though, and get these 2 packages to work against CFN.

This template snippet demonstrates usage (you need your own code packages and S3 object links - make sure the code is actually different in each package or you'll eventually run into a separate bug which we need to fix);

```
Resources:
    InitialType:
        Type: AWS::CloudFormation::Type
        Properties:
            Type: RESOURCE
            TypeName: Sample::DadJoke::Resource
            SchemaHandlerPackage: s3://cloudformationmanageduploadinfrast-artifactbucket-123456789012abcdef/sample-cloudformation-resource.zip
    UpdatedType:
        Type: AWS::CloudFormation::Type
        Properties:
            Type: RESOURCE
            TypeName: Sample::DadJoke::Resource
            SchemaHandlerPackage: s3://cloudformationmanageduploadinfrast-artifactbucket-123456789012abcdef/sample-cloudformation-resource-update.zip
        DependsOn: InitialType

    DefaultVersion:
        Type: AWS::CloudFormation::TypeVersion
        Properties:
            Arn: !Ref UpdatedType
           #  DefaultVersion: !GetAtt UpdatedType.VersionId
            DefaultVersion: 00000002
```

benkehoe:
I like this resource layout, but I think the names are wrong. I'd propose AWS::CloudFormation::TypeVersion and AWS::CloudFormation::TypeVersionAlias. So it would look like this:

```yaml
Resources:
    InitialType:
        Type: AWS::CloudFormation::TypeVersion
        Properties:
            Type: RESOURCE
            TypeName: Sample::CloudFormation::Resource
            SchemaHandlerPackage: s3://cloudformationmanageduploadinfrast-artifactbucket-123456789012abcdef/sample-cloudformation-resource.zip
    UpdatedType:
        Type: AWS::CloudFormation::TypeVersion
        Properties:
            Type: RESOURCE
            TypeName: Sample::CloudFormation::Resource
            SchemaHandlerPackage: s3://cloudformationmanageduploadinfrast-artifactbucket-123456789012abcdef/sample-cloudformation-resource-update.zip
        DependsOn: InitialType

    DefaultVersion:
        Type: AWS::CloudFormation::TypeVersionAlias
        Properties:
            Arn: !Ref UpdatedType
            AliasName: $default # only valid value currently
            # VersionId: !GetAtt UpdatedType.VersionId
            VersionId: 00000002
```
I'd argue for a) having an alias name that required to be a specific value, rather than making default version different from possible future aliases _from the IaC perspective_ (templates don't need to match the API calls) and b) going with  `$default` as the alias name as that's what Lambda and API Gateway are using in their alias naming.
As to VersionId: is it a string or a number? Leading zeros are hard in YAML (see @rhboyd's tribulations with account IDs in mappings).

I'm sad that it's `AWS::CloudFormation::TypeVersion`+`Type=RESOURCE` and not `AWS::CloudFormation::ResourceVersion`. I get that there will be different types in the future, but I expect that they will be managed by users differently, have different lifecycles, etc., so I having been hoping that they could be different CloudFormation types. If they were separate CloudFormation types, when you release a new registry type, say it's called "Foo", a generic CloudFormation analyzer would know "this template defines M new Resources and N new Foos and K IAM roles and..." without knowing anything about the properties of the resources. Under this scheme, it would only be able to say "this template defines X CloudFormation types 🤷‍♂️ and K IAM roles and...", and while X=M+N, it would take additional specialized code to split that out.

I get that this will save duplication of code between the resources and matches the API, but it feels like optimizing this handler implementation at the expense of the user experience.

rjlohan:
> I'm sad that it's `AWS::CloudFormation::TypeVersion`+`Type=RESOURCE` and not `AWS::CloudFormation::ResourceVersion`.

I could probably be convinced that this is OK. You're right that I'm optimizing for implementation detail here and that's probably the wrong approach. Can always share 99% of the code behind slightly different schemas if needed. There's a bit of YAGNI here too and I should worry about what we have more than what we may have later. 😄

benkehoe:
Oh, fabulous! So you can be convinced--what do I need to do to convince you?

rjlohan:
> Oh, fabulous! So you can be convinced--what do I need to do to convince you?

Send beer. 👌

benkehoe:
I _tried_ to buy you all beer at re:Invent for exactly this kind of eventuality but Amjad wouldn't let me

rjlohan:
Latest revision has the new type names as suggested. Still a problem on setting the `DefaultVersionId` that I have to fix up.

benkehoe:
Pinging on my question about `DefaultVersion: 00000002` in your resource example. I have a hard time understanding what an eventual non-default alias would look like, since it must presumably have a name and `VersionId`, so I imagine something like:

```yaml
MyVersion:
    Type: AWS::CloudFormation::TypeVersionAlias
    Properties:
        Arn: !Ref UpdatedType
        AliasName: SomeAliasName
        # VersionId: !GetAtt UpdatedType.VersionId
        VersionId: 00000002
```

Given that, it seems like the `VersionId` field should be the same whether it's a default or "custom" alias. So then the default alias could either reuse `AliasName` with a special value like `$default` (what Lambda and API Gateway use), or something like `IsDefaultAlias: true`

rjlohan:
> Pinging on my question about `DefaultVersion: 00000002` in your resource example. I have a hard time understanding what an eventual non-default alias would look like, since it must presumably have a name and `VersionId`, so I imagine something like:
> 
> ```yaml
> MyVersion:
>     Type: AWS::CloudFormation::TypeVersionAlias
>     Properties:
>         Arn: !Ref UpdatedType
>         AliasName: SomeAliasName
>         # VersionId: !GetAtt UpdatedType.VersionId
>         VersionId: 00000002
> ```
> 
> Given that, it seems like the `VersionId` field should be the same whether it's a default or "custom" alias. So then the default alias could either reuse `AliasName` with a special value like `$default` (what Lambda and API Gateway use), or something like `IsDefaultAlias: true`

Yeah that makes sense. I'm still fixated on the current nomenclature having only a `Default`. I'll probably rework this - consistency with other similar types like `AWS::Lambda::Alias` is a good way to go.

Right now I'm blocked on completion by a change I need to make on the backend - which I've done, just needs to be rolled out.

ikben:
I'm not sure if this is already provided by the code/rpdk (my java is very rusty), but to come back on:
```yaml
DefaultVersion: !GetAtt UpdatedType.VersionId
DefaultVersion: "00000002"
DefaultVersion: 00000002
```
I think it might make sense to actually fail and give an error message in the case that the supplied version is a number (even if the API would be able to set the default version.

- I expect that in most cases the DefaultVersion will be a GetAtt or a Ref to a Parameter, in which case it doesn't matter
- In the case that it's not it will work fine for 00000001 to 00000009, but suddenly fail on 00000010 (or worse, revert to the old version, "00000008" if it sill exists). Which will lead to frustration and/or bugs
- Throwing errors on numbers will be annoying the first time someone runs into it, but I think its preferable to the alternative

benkehoe:
It's definitely unclear whether version numbers are strings or numbers, and whether or not the leading zeros are required. If they are...are you sure you have enough? I might need 100 million versions, I am constantly screwing things up 😜

rhboyd:
The string/number uncertainty would be consistent with the rest of the cfn service though

ikben:
> The string/number uncertainty would be consistent with the rest of the cfn
> service though
>
>
Almost consistent. IIRC Elastic Beanstalk Option Settings fail if the
values are not strings.

pgarbe:
Did I get it right that the type `AWS::CloudFormation::ResourceVersion` is immutable and can't be changed after it's created. So, if I want to provide a newer implementation I've to create a second resource in my template and change the default version. Once that is deployed, I can remove the first version, right? 

I'm not sure if makes sense to compare it with Lambda Version/Alias. A typical use case that I've in mind is to just change the package version that is provided by the Vendor. So my first deployment is:

```yaml
Resources:
    MySampleResource:
        Type: AWS::CloudFormation::Resource
        Properties:
            TypeName: Sample::CloudFormation::Resource
            SchemaHandlerPackage: s3://cloudformationmanageduploadinfrast-artifactbucket-123456789012abcdef/sample-cloudformation-resource-v0.0.1.zip
```

and in the second deployment, I just change the package location:

```yaml
Resources:
    MySampleResource:
        Type: AWS::CloudFormation::Resource
        Properties:
            TypeName: Sample::CloudFormation::Resource
            SchemaHandlerPackage: s3://cloudformationmanageduploadinfrast-artifactbucket-123456789012abcdef/sample-cloudformation-resource-v0.0.2.zip
```

Everything else should be handled by CloudFormation and as user I don't care how many and which API calls are necessary. And in that case I wouldn't call it `ResourceVersion` but just `Resource`.

benkehoe:
You can do exactly what you are describing with `AWS::CloudFormation::ResourceVersion`. The semantics are that updates to `SchemaHandlerPackage` require replacement, so when you've changed that property, a new version will be created and the old version deleted. The reason why it's `::ResourceVersion` is that it represents a specific _version_ of the resource type, rather than the resource type as a whole (note that if you `!Ref` it, it would be an ARN to the version), and in particular, you can have more than one `::ResourceVersion` of the same resource type in the same template (for example, if you're supporting a "v1.x" for one team and "v2.x" for another team). This is something that isn't possible with Lambda. For example, you can't deploy two versions of a Lambda function and a weighted alias with references to both in a single template. There's a bit more explanation on this topic in [this article](https://read.acloud.guru/cloudformation-is-an-infrastructure-graph-management-service-and-needs-to-act-more-like-it-fa234e567c82).

pgarbe:
Thanks @benkehoe for clarification. So both use cases are supported which is perfect. Now it just needs to be shipped...

benkehoe:
We've still got some questions about `::ResourceVersionAlias` to work out:
- How is the default alias represented?
- Are version ids numbers or strings (or both)?

rjlohan:
> * How is the default alias represented?

I'll come back to this when I get the SDK update for the missing field.

> * Are version ids numbers or strings (or both)?

They are strings representing for 8 digit incrementing integers. 🤷‍♂

benkehoe:
> They are strings representing for 8 digit incrementing integers. 🤷‍♂

I guess what I mean is, is the zero padding is required or optional? The [`SetTypeDefaultVersion` API](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_SetTypeDefaultVersion.html) specifies the pattern to be `[A-Za-z0-9-]+`, which is of no help as letters are clearly not valid at present.

rjlohan:
> > They are strings representing for 8 digit incrementing integers. 🤷‍♂
> 
> I guess what I mean is, is the zero padding is required or optional? The [`SetTypeDefaultVersion` API](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_SetTypeDefaultVersion.html) specifies the pattern to be `[A-Za-z0-9-]+`, which is of no help as letters are clearly not valid at present.

The version string should be treated as a string, as we reserve the right to change their format in the future. So, the "0" padding is required because they're not numbers, they're strings.

benkehoe:
@rjlohan and I talked about this a few weeks ago. I came around to the idea that `::ResourceDefaultVersion` should be separate from a future `::ResourceVersionAlias`. In particular, `::ResourceVersionAlias` would represent something like "1.x" versus "2.x" and would point to actual versions, but the `::ResourceDefaultVersion` could point to a resource version alias (so you'd be able to say that by default people should be on 2.x). I do think the field should be `VersionId`, though. So:
```yaml
    DefaultVersion:
        Type: AWS::CloudFormation::ResourceDefaultVersion
        Properties:
            Arn: !Ref UpdatedType
            #  VersionId: !GetAtt UpdatedType.VersionId
            VersionId: 00000002
```

Future version aliases could look like this:
```yaml
    VersionAlias:
        Type: AWS::CloudFormation::ResourceVersionAlias
        Properties:
            Arn: !Ref UpdatedType
            AliasName: "2.x"
            #  VersionId: !GetAtt UpdatedType.VersionId
            VersionId: 00000002

    DefaultVersion:
        Type: AWS::CloudFormation::ResourceDefaultVersion
        Properties:
            Arn: !Ref UpdatedType
            VersionAlias: !Ref VersionAlias
```

rjlohan:
> @rjlohan and I talked about this a few weeks ago. I came around to the idea that `::ResourceDefaultVersion` should be separate from a future `::ResourceVersionAlias`. In particular, `::ResourceVersionAlias` would represent something like "1.x" versus "2.x" and would point to actual versions, but the `::ResourceDefaultVersion` could point to a resource version alias (so you'd be able to say that by default people should be on 2.x). I do think the field should be `VersionId`, though. So:
> 
> ```yaml
>     DefaultVersion:
>         Type: AWS::CloudFormation::ResourceDefaultVersion
>         Properties:
>             Arn: !Ref UpdatedType
>             #  VersionId: !GetAtt UpdatedType.VersionId
>             VersionId: 00000002
> ```
> 
> Future version aliases could look like this:
> 
> ```yaml
>     VersionAlias:
>         Type: AWS::CloudFormation::ResourceVersionAlias
>         Properties:
>             Arn: !Ref UpdatedType
>             AliasName: "2.x"
>             #  VersionId: !GetAtt UpdatedType.VersionId
>             VersionId: 00000002
> 
>     DefaultVersion:
>         Type: AWS::CloudFormation::ResourceDefaultVersion
>         Properties:
>             Arn: !Ref UpdatedType
>             VersionAlias: !Ref VersionAlias
> ```

Thanks for writing that up, I would have forgotten. :-/

pgarbe:
Any updates here?

wolfeidau:
Just started using resource providers and really keen on seeing this PR progress, is there anything we can do to help?

richardhboyd:
One capability that I’d like to see in this is being able to register a new type and use it in the same template. Current validation mechanisms would fail because the type is unrecognized when stack creation starts.

rjlohan:
> One capability that I’d like to see in this is being able to register a new type and use it in the same template. Current validation mechanisms would fail because the type is unrecognized when stack creation starts.

Not a bad idea, but it's probably not something that can be achieved *by* this implementation. As you note - for the first registration the type won't exist when CloudFormation validates the existence of types up front. And beyond that, while the type would exist at that point, CloudFormation would use the previously registered version for the stack operation, rather than fetch this new one.

I don't have a clear picture on how that request might be solved without breaking this validation flow, but open to suggestions.

rjlohan:
> Just started using resource providers and really keen on seeing this PR progress, is there anything we can do to help?

Thanks but it's really just waiting on an SDK update with a required property, which is in progress.

ikben:
> One capability that I’d like to see in this is being able to register a new type and use it in the same template. Current validation mechanisms would fail because the type is unrecognized when stack creation starts.

I understand that need, but I personally consider that an anti-pattern when used with Custom Resources, so I would be careful when adding this. 

Macros (inherently) have the same limitation, and I believe that's the better model to follow everywhere (no template can change it's own runtime behaviour).
This does raise an interesting question about what will happen if you switch versions while a template is in progress.

rjlohan:
> This does raise an interesting question about what will happen if you switch versions while a template is in progress.

When a stack operation starts, CloudFormation grabs the default version of all resources used in the template and 'pins' them for the remainder of that stack operation. This is how type versioning has always been handled in CloudFormation, albeit with different implementation details.

richardhboyd:
The version is frozen when the stack operation starts? Does that mean various nested templates might use a different versions of a type of that type is updated out-of-band (by the CLI or another CFN Template)?

rjlohan:
> The version is frozen when the stack operation starts? Does that mean various nested templates might use a different versions of a type of that type is updated out-of-band (by the CLI or another CFN Template)?

No, they should use the same version too as the version for the type is already captured up front. The only time distortion you might see is that if a nested template contains a type that hasn't been used yet, it might use a version that was registered after the stack operation started.

richardhboyd:
Suppose I have 3 stacks, one root (StackA) and two sibling nested stacks (StackB and StackC). StackA doesn't use `MyType` at all, but StackB and StackC both use it. (Let's also suppose StackC has an explicit DependsOn for StackB). Are you saying that when StackB uses `MyType` then the version is locked and StackC will use the same version?

richardhboyd:
oof, I just realized I added my comment on the PR instead of the issue. feel free to ignore my comments here.

rjlohan:
> Suppose I have 3 stacks, one root (StackA) and two sibling nested stacks (StackB and StackC). StackA doesn't use `MyType` at all, but StackB and StackC both use it. (Let's also suppose StackC has an explicit DependsOn for StackB). Are you saying that when StackB uses `MyType` then the version is locked and StackC will use the same version?

That's how it should work, yes.

rjlohan:
> > Just started using resource providers and really keen on seeing this PR progress, is there anything we can do to help?
> 
> Thanks but it's really just waiting on an SDK update with a required property, which is in progress.

Guess I'm out of excuses now: https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/pull/267

:)

rjlohan:
@benkehoe @richardhboyd @ikben @pgarbe @wolfeidau I have incorporated the SDK updates and done some basic end-to-end testing. If you are willing and able to provide some usabililty feedback or spot bugs, the current code can be cloned/forked and provisioned as a private type.

Appreciate any feedback you can give while I polish the tests and what not.

benkehoe:
It looks like these changes aren't present yet: https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-cloudformation/pull/4#issuecomment-603902255

rjlohan:
> It looks like these changes aren't present yet: [#4 (comment)](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-cloudformation/pull/4#issuecomment-603902255)

No, but imagine `ResourceTypeAlias` is renamed to `ResourceDefaultAlias` and the inline property is just part of the build. I can do that later if the usability pans out correctly.

There's no such thing as a named alias in the service so there's no way to implement this at the moment, except as a template construct with no backing resource.

aws-cloudformation-resourceversion/src/main/java/software/amazon/cloudformation/resourceversion/ReadHandler.java
```diff
@@ -0,0 +1,86 @@
...
+            throw nullSafeNotFoundException(model);
+        } catch (final CfnRegistryException e) {
+            logger.log(Arrays.toString(e.getStackTrace()));
+            throw new CfnGeneralServiceException(e);
+        }
+
+        final ResourceModel modelFromReadResult = Translator.translateForRead(response);
+
+        // set the version and defaultVersion from the model ARN, not the Read response, which
+        // will use the Type details, rather than the specific version details
```
aygold92, line 66: if you pass the versionId into the describe-type call, it will return details specific to that version. I assumed it would do that if you pass the full versioned ARN as well

rjlohan:
Don't stress too much over a code review here, I have to do some non-logic-affecting rewrites based on latest Java Plugin updates, just want some usability feedback on the types.

aws-cloudformation-resourceversionalias/aws-cloudformation-resourceversionalias.json
```diff
@@ -0,0 +1,74 @@
...
+      ]
+    },
+    "update": {
+      "permissions": [
+        "cloudformation:SetTypeDefaultVersion"
+      ]
+    },
+    "delete": {
+      "permissions": [
+        ""
```
aygold92, line 69: definitely don't like that the delete action does nothing. Setting the default version feels more like a hook action right?  What benefit does having it as a resource provide?

aws-cloudformation-resourceversion/aws-cloudformation-resourceversion.json
```diff
@@ -0,0 +1,160 @@
...
+      ],
+      "type": "string"
+    }
+  },
+  "required": [
+    "SchemaHandlerPackage",
+    "TypeName"
+  ],
+  "readOnlyProperties": [
+    "/properties/Arn",
```
jlhood, line 116: Can you add `/properties/TypeArn` as well so you can get the ARN for the type without the version id attached?

rjlohan: I _could_, though that isn't strictly a property of the Resource*Version*. It's also not returned out of the `DescribeType` API, so I'd be stripping the version from the Arn. What is the use-case for this you think?

aws-cloudformation-resourceversion/src/main/java/software/amazon/cloudformation/resourceversion/CreateHandler.java
```diff
@@ -0,0 +1,173 @@
...
+    ProgressEvent<ResourceModel, CallbackContext> predictArn(
+        final AmazonWebServicesClientProxy proxy,
+        final ResourceHandlerRequest<ResourceModel> request,
+        final CallbackContext callbackContext,
+        final Logger logger) {
+
+        String arn;
+        try {
+            // find the highest existing version number for this type
+            while (true) {
```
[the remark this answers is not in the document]
rjlohan: To be more resilient to arbitrary ordering I should really enumerate until I find the latest.

aws-cloudformation-resourceversion/src/main/java/software/amazon/cloudformation/resourceversion/ClientBuilder.java
```diff
@@ -0,0 +1,10 @@
+package software.amazon.cloudformation.resourceversion;
+
+import software.amazon.awssdk.services.cloudformation.CloudFormationClient;
+
+public class ClientBuilder {
+    public static CloudFormationClient getClient() {
+        return CloudFormationClient.builder()
```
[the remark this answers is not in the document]
rjlohan: The SDK clients already have a `defaultRetryPolicy` applied, that retries on all of these conditions: https://github.com/aws/aws-sdk-java-v2/blob/e971765bb6be61e3fb510357154279c555f7e521/core/aws-core/src/main/java/software/amazon/awssdk/awscore/internal/AwsErrorCode.java#L30

eduardomourar:
Could we also add a `SemanticVersion` property like SAR does (docs [here](https://docs.aws.amazon.com/serverlessrepo/latest/devguide/applications-applicationid-versions-semanticversion.html))? That way we can ensure we are not uploading the same version over and over again during update event.

aws-cloudformation-resourceversion/aws-cloudformation-resourceversion.json
eduardomourar: I believe this is important because it is the only unique identifier for the resource type (it has the name, type and region in one place). The response from `DescribeTypeRegistration` API  does give that, right?

richardhboyd: This ARN would have that info.

rjlohan: The Version Arn carries the Type Arn, but the Version is always there. Essentially, calling `DescribeType` *without* specifying a version implicitly describes the `default` version. As a Type owner, you *always* get back a Version Arn. For types you do not own, you *never* get back a Version Arn.

rjlohan: So, essentially, if you are provisioning types, you are always the owner, and you should always think about them in terms of a Version Arn. A type cannot exist without a version.

aws-cloudformation-resourceversion/src/main/java/software/amazon/cloudformation/resourceversion/CreateHandler.java
```diff
@@ -0,0 +1,136 @@
...
+        return proxy.initiate("AWS-CloudFormation-ResourceVersion::Create", proxyClient, resourceModel, context)
+            .translateToServiceRequest(Translator::translateToCreateRequest)
+            .makeServiceCall((awsRequest, sdkProxyClient) -> sdkProxyClient.injectCredentialsAndInvokeV2(awsRequest, sdkProxyClient.client()::registerType))
+            .done((registerTypeRequest, registerTypeResponse, sdkProxyClient, model, cc) -> {
+                cc.setRegistrationToken(registerTypeResponse.registrationToken());
+                return predictArn(sdkProxyClient, request, cc);
+            })
+            .then(progress -> stabilizeOnCreate(proxy, proxyClient, progress))
+            .then(progress -> new ReadHandler().handleRequest(proxy, request, context, proxyClient, logger));
+    }
```
ammokhov, line 46: ideally u dont need to put any additional vars in callbackContext;
Id reworked it like that:
```java
        return proxy.initiate("AWS-CloudFormation-ResourceVersion::Create", proxyClient, resourceModel, context)
            .translateToServiceRequest(Translator::translateToCreateRequest)
            .makeServiceCall((awsRequest, sdkProxyClient) -> sdkProxyClient.injectCredentialsAndInvokeV2(awsRequest, sdkProxyClient.client()::registerType))
            .stabilize((awsRequest, awsResponse, client, model, cc) -> {
                if (model.getArn() == null) { 
                    model.setArn(predictArn(awsResponse..registrationToken()) 
                }
                final describeResponse = sdkProxyClient.injectCredentialsAndInvokeV2(awsRequest, sdkProxyClient.client()::describeTypeRegistration);       
                

                if (response.progressStatus().equals(RegistrationStatus.COMPLETE)) {
                    logger.log(String.format("%s registration successfully completed [%s].", ResourceModel.TYPE_NAME, response.typeVersionArn()));
                    return true;
                } else if (response.progressStatus().equals(RegistrationStatus.FAILED)) {
                    logger.log(String.format("Registration request %s failed with '%s'", registrationToken, response.description()));
                    throw new CfnNotStabilizedException(ResourceModel.TYPE_NAME, resourceModel.getArn());
                } else {
                    return false;
                }
 
            })
            .progress()
            .then(progress -> new ReadHandler().handleRequest(proxy, request, context, proxyClient, logger));

```

aws-cloudformation-resourceversion/src/main/java/software/amazon/cloudformation/resourceversion/DeleteHandler.java
```diff
@@ -0,0 +1,40 @@
...
+
+        // pre-read to capture required metadata fields in model for Delete
+        return new ReadHandler().handleRequest(proxy, request, callbackContext, proxyClient, logger)
+            // now deregister the type
+            .onSuccess(progress ->
+                proxy.initiate("AWS-CloudFormation-ResourceVersion::Delete", proxyClient, progress.getResourceModel(), callbackContext)
+                    .translateToServiceRequest(awsRequest ->
+                        Translator.translateToDeleteRequest(awsRequest, logger))
+                    .makeServiceCall((awsRequest, proxyInvocation) -> proxyInvocation.injectCredentialsAndInvokeV2(awsRequest, proxyInvocation.client()::deregisterType))
+                    .success());
```
ammokhov, line 32: that's pretty cool

aws-cloudformation-resourceversion/src/main/java/software/amazon/cloudformation/resourceversion/ReadHandler.java
```diff
@@ -0,0 +1,95 @@
...
+        // set the version from the model ARN, not the Read response, which
+        // will use the Type details, rather than the specific version details
+        final String versionId = model.getArn().substring(model.getArn().lastIndexOf('/') + 1);
+        modelFromReadResult.setArn(model.getArn());
+        modelFromReadResult.setVersionId(versionId);
+
+        return ProgressEvent.defaultSuccessHandler(modelFromReadResult);
+    }
+
+    private software.amazon.cloudformation.exceptions.ResourceNotFoundException nullSafeNotFoundException(final ResourceModel model) {
```
ammokhov, line 90: make sense to put that in `BaseHandlerStd` if that's used by multiple handlers

rjlohan: Think I'll end up dropping this after another iteration.

aws-cloudformation-resourceversion/src/main/java/software/amazon/cloudformation/resourceversion/CreateHandler.java
```diff
@@ -0,0 +1,136 @@
...
+        final AmazonWebServicesClientProxy proxy,
+        final ResourceHandlerRequest<ResourceModel> request,
+        final CallbackContext callbackContext,
+        final ProxyClient<CloudFormationClient> proxyClient,
+        final Logger logger) {
+
+        this.logger = logger;
+
+        final ResourceModel resourceModel = request.getDesiredResourceState();
+        final CallbackContext context = callbackContext == null ? CallbackContext.builder().build() : callbackContext;
```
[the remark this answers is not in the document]
rjlohan: Ah that's right...!

aws-cloudformation-resourceversion/src/main/java/software/amazon/cloudformation/resourceversion/ReadHandler.java
rjlohan: See comment below about how to understand type/version Arns here.

aws-cloudformation-resourceversion/src/main/java/software/amazon/cloudformation/resourceversion/ListHandler.java
```diff
@@ -0,0 +1,29 @@
...
+    @Override
+    public ProgressEvent<ResourceModel, CallbackContext> handleRequest(
+        final AmazonWebServicesClientProxy proxy,
+        final ResourceHandlerRequest<ResourceModel> request,
+        final CallbackContext callbackContext,
+        final Logger logger) {
+
+        final ListTypesResponse response =
+            proxy.injectCredentialsAndInvokeV2(Translator.translateToListRequest(request.getNextToken()),
+                ClientBuilder.getClient()::listTypes);
```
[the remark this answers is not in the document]
rjlohan: That... is interesting. You cannot list all type versions on the API - you have to supply an Arn or TypeName to filter by. Which breaks the List contract. So.... either we list all Types or enumerate all Types and then List all versions of each... the latter is probably most correct.

aws-cloudformation-resourceversion/aws-cloudformation-resourceversion.json
```diff
@@ -0,0 +1,160 @@
...
+          "pattern": "^arn:.+:iam::[0-9]{12}:role/.+$",
+          "minLength": 1,
+          "maxLength": 256
+        }
+      },
+      "additionalProperties": false
+    }
+  },
+  "properties": {
+    "Arn": {
```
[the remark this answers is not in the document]
rjlohan: Conversation below.

rjlohan:
Updated the implementation so that example usage looks like;


```
Resources:
    InitialType:
        Type: AWS::CloudFormation::ResourceVersion
        Properties:
            TypeName: Sample::CloudFormation::Resource
            SchemaHandlerPackage: s3://cloudformationmanageduploadinfrast-artifactbucket-123456789012abcdef/sample-cloudformation-resource.zip
    UpdatedType:
        Type: AWS::CloudFormation::ResourceVersion
        Properties:
            TypeName: Sample::CloudFormation::Resource
            SchemaHandlerPackage: s3://cloudformationmanageduploadinfrast-artifactbucket-123456789012abcdef/sample-cloudformation-resource-update.zip
        DependsOn: InitialType

    DefaultVersion:
        Type: AWS::CloudFormation::ResourceDefaultVersion
        Properties:
            Arn: !Ref UpdatedType
```

rjlohan:
I'm going to work through some additional testing on this, but I'm settled on this model now unless there's other critical feedback.

benkehoe:
> `Arn: !Ref UpdatedType`

Ohh, brilliant. Because there's no separate `Resource` to reference, why should there be separate `Type` and `VersionId` fields, you're just referencing the `ResourceVersion`. I love it.

Is the `DependsOn` necessary because there might be cross-talk if they get created at the same time? Otherwise, it shouldn't matter which one gets created first, right? Is this something https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/issues/79 would help with?

rjlohan:
> > `Arn: !Ref UpdatedType`
> 
> Ohh, brilliant. Because there's no separate `Resource` to reference, why should there be separate `Type` and `VersionId` fields, you're just referencing the `ResourceVersion`. I love it.

Yep, but I've allowed both options in case you're managing across stacks or something. I'd expect `!Ref UpdatedType` to be the usual usage though.

> Is the `DependsOn` necessary because there might be cross-talk if they get created at the same time? Otherwise, it shouldn't matter which one gets created first, right? Is this something [aws-cloudformation/aws-cloudformation-resource-schema#79](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/issues/79) would help with?

The Registry restricts only a single in-flight `registerType` action for a type, so this ordering ensures that we don't attempt to register both at once as that will fail.

ikben:
> > The Registry restricts only a single in-flight `registerType` action for a type, so this ordering ensures that we don't attempt to register both at once as that will fail.
> 
> Right. This is off topic, but it seems useful if you could put something like `"serializedProperties": [ "TypeName" ]` in your `::ResourceVersion` schema and CloudFormation would make sure it's only running one operation at a time for `::ResourceVersion` resources with a common `TypeName` (of `Sample::CloudFormation::Resource` in this case)

Even more off-topic: it would be even better if it could accept a number (eg. for dynamodb tables where there can be 10 creates at the same time)

rjlohan:
> > > The Registry restricts only a single in-flight `registerType` action for a type, so this ordering ensures that we don't attempt to register both at once as that will fail.
> > 
> > 
> > Right. This is off topic, but it seems useful if you could put something like `"serializedProperties": [ "TypeName" ]` in your `::ResourceVersion` schema and CloudFormation would make sure it's only running one operation at a time for `::ResourceVersion` resources with a common `TypeName` (of `Sample::CloudFormation::Resource` in this case)
> 
> Even more off-topic: it would be even better if it could accept a number (eg. for dynamodb tables where there can be 10 creates at the same time)

Have you started this ask on the roadmap yet? Let's take it there. What you're really asking for is a customer-driven throttling control. I can see value,  but needs more deep dive and I can't speak to priority.

benkehoe:
I opened it on the schema repo here: https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/issues/79 . Should I open it on the roadmap instead?

rjlohan:
> I opened it on the schema repo here: [aws-cloudformation/aws-cloudformation-resource-schema#79](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/issues/79) . Should I open it on the roadmap instead?

No that's OK.

rjlohan:
I *believe* this version of the PR to be finalised. It depends on 2 pending changes in the Java plugin repo if testing;
* https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/pull/287
* https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/pull/288

aws-cloudformation-resourceversion/src/main/java/software/amazon/cloudformation/resourceversion/CreateHandler.java
```diff
@@ -0,0 +1,181 @@
...
+                                    ListTypeVersionsResponse.builder().typeVersionSummaries(
+                                        getPredictedSummary(request, model)
+                                    ).build(),
+                                    context
+                                );
+                            }
+                            throw exception;
+                        }
+                    )
+                    .done(response_ -> ProgressEvent.success(response_, context));
```
ammokhov, line 83: maybe just use a direct api call as u dont really register anything in the callgraph but it makes code less verbose

aws-cloudformation-resourceversion/src/main/java/software/amazon/cloudformation/resourceversion/CreateHandler.java
```diff
@@ -0,0 +1,181 @@
...
+        final CallChain.Initiator<CloudFormationClient, ResourceModel, CallbackContext> initiator =
+            proxy.newInitiator(proxyClient, resourceModel, callbackContext);
+
+        return initiator
+            .translateToServiceRequest(Translator::translateToCreateRequest)
+            .makeServiceCall((awsRequest, sdkProxyClient) -> sdkProxyClient.injectCredentialsAndInvokeV2(awsRequest, sdkProxyClient.client()::registerType))
+            .done((registerTypeRequest, registerTypeResponse, sdkProxyClient, model, cc) -> {
+                cc.setRegistrationToken(registerTypeResponse.registrationToken());
+                return predictArn(request, initiator);
+            })
```
ammokhov, line 39: just a thought but can we make registration token a primary id? we dont need to predict in this case

rjlohan:
Contract tests for `AWS::CloudFormation::ResourceVersion` are now functional with 2 exceptions;

* `contract_read_without_create` fails due to https://github.com/aws-cloudformation/cloudformation-cli/issues/507 though is not actually a problem in this resource type
* `contract_create_list_success` fails because the list handler doesn't return type *versions*, I have to look at this

```
=========================================================================================================== test session starts ============================================================================================================
platform darwin -- Python 3.7.5, pytest-5.4.1, py-1.8.0, pluggy-0.13.1 -- /Users/ryanloha/workspaces/cloudformation-cli/env/bin/python3
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/Users/ryanloha/workspaces/aws-cloudformation-resource-providers-cloudformation/aws-cloudformation-resourceversion/.hypothesis/examples')
Test order randomisation NOT enabled. Enable with --random-order or --random-order-bucket=<bucket_type>
rootdir: /Users/ryanloha/workspaces/aws-cloudformation-resource-providers-cloudformation/aws-cloudformation-resourceversion, inifile: /private/var/folders/w8/jwm15b8n64l4gwc9ctjs4w4xj27kh4/T/pytest_eunwx3t0.ini
plugins: hypothesis-4.53.2, localserver-0.5.0, random-order-1.0.4, cov-2.8.1
collected 16 items / 5 deselected / 11 selected                                                                                                                                                                                            

handler_create.py::contract_create_delete PASSED                                                                                                                                                                                     [  9%]
handler_create.py::contract_invalid_create PASSED                                                                                                                                                                                    [ 18%]
handler_create.py::contract_create_duplicate PASSED                                                                                                                                                                                  [ 27%]
handler_create.py::contract_create_read_success PASSED                                                                                                                                                                               [ 36%]
handler_create.py::contract_create_list_success FAILED                                                                                                                                                                               [ 45%]
handler_delete.py::contract_delete_read PASSED                                                                                                                                                                                       [ 54%]
handler_delete.py::contract_delete_list PASSED                                                                                                                                                                                       [ 63%]
handler_delete.py::contract_delete_delete PASSED                                                                                                                                                                                     [ 72%]
handler_delete.py::contract_delete_create SKIPPED                                                                                                                                                                                    [ 81%]
handler_misc.py::contract_check_asserts_work PASSED                                                                                                                                                                                  [ 90%]
handler_read.py::contract_read_without_create FAILED                                                                                                                                                                                 [100%]


```

aygold92:
lets get this out and we can iterate on it

.travis.yml
```diff
@@ -10,6 +10,20 @@ install:
...
+    -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
+    -B
+    clean verify
+  - cd "$TRAVIS_BUILD_DIR/aws-cloudformation-resourceversion"
+  # from Maven 3.6.1+, should use `--no-transfer-progress` instead of Slf4jMavenTransferListener
+  - >
+    mvn
+    -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
+    -B
+    clean verify
```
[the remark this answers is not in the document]
miparnisari: https://github.com/aws-cloudformation/devtools/pull/11

PR #6, opened 2020-04-22
Add Stackset Resource

xiwhuang:
### Description of changes:
Added new resource AWS::CloudFormation::StackSet

### Lastest Revision:
* Refactored handler implementations using Call Chaining Pattern.
* Deleted `AwsCredentialsExtrator` as it is no longer needed.
* Updated to use GetTemplateSummary to validate template.

#### Unit Tests
* 97% Instructions Cov.
* 92% Branches Cov.
#### Manually tests
* StackSet without StackInstances `CREATE`/`DELETE`/`UPDATE`
* StackSet with StackInstances `CREATE`/`DELETE`/`UPDATE`
* StackSet CREATE/DELETE/UPDATE with `TemplateURL`/`TemplateBody`
* Negative Tests with `Invalid Template`
  * Unsupported regions
  * Embedded Stack/StackSet
  * Invalid `TemplateURL`/ `TemplateBody`
* Template with Outputs

#### Known limitation
* `TemplateURL` and `OrganizationUnitIds` can be not retrieve in READ/LIST handlers as StackSet API currently does not allow to do so.

### Simple Working Template
```yaml
Resources:
  MyStackSet:
    Type: AWS::CloudFormation::StackSet
    Properties:
      PermissionModel: SELF_MANAGED
      TemplateBody: |
        {
          "AWSTemplateFormatVersion": "2010-09-09",
          "Resources": {
            "IntegrationTestWaitHandle": {
              "Type": "AWS::CloudFormation::WaitConditionHandle",
              "Properties": {}
            }
          }
        }
      StackInstancesGroup: 
        - Regions:
            - us-east-1
            - us-west-2
          DeploymentTargets:
            Accounts:
              - 111111111111
      Tags:
        - Key: key1
          Value: value1
Outputs:
  StackSetId:
    Value: !Ref MyStackSet
```


aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/CreateHandler.java
```diff
@@ -0,0 +1,85 @@
...
+        final ResourceModel model = request.getDesiredResourceState();
+        final String stackSetName = PhysicalIdGenerator.generatePhysicalId(request);
+        analyzeTemplate(proxy, model, callbackContext);
+
+        return proxy.initiate("AWS-CloudFormation-StackSet::Create", proxyClient, model, callbackContext)
+                .request(resourceModel -> createStackSetRequest(resourceModel, stackSetName, request.getClientRequestToken()))
+                .call((modelRequest, proxyInvocation) -> createResource(modelRequest, proxyClient, model))
+                .progress()
+                .then(progress -> createStackInstances(proxy, proxyClient, progress, logger))
+                .then(progress -> ProgressEvent.defaultSuccessHandler(model));
```
xiwhuang, line 43:
```java
return proxy.initiate("AWS-CloudFormation-StackSet::Create", proxyClient, model, callbackContext)
     .request(resourceModel -> createStackSetRequest(resourceModel, stackSetName, request.getClientRequestToken()))
     .retry(MULTIPLE_OF)
     .call((modelRequest, proxyInvocation) -> createResource(modelRequest, proxyClient, model))
     .progress()
     .then(progress -> createStackInstances(proxy, proxyClient, progress, logger))
     .then(progress -> {
          if (progress.isFailed()) return progress;
          return ProgressEvent.defaultSuccessHandler(model);
      });
```
Trying to do something like above, not sure how to implement it. Above implementation failed, meaning it will still return success no matter what.

aws-cloudformation-stackset/aws-cloudformation-stackset.json
```diff
@@ -0,0 +1,309 @@
...
+            "maxLength": 1024
+        }
+    },
+    "required": [
+        "PermissionModel"
+    ],
+    "additionalProperties": false,
+    "createOnlyProperties": [
+        "/properties/PermissionModel"
+    ],
```
PatMyron, line 252:
```suggestion
```

[`PermissionModel` can be updated](https://docs.aws.amazon.com/cli/latest/reference/cloudformation/update-stack-set.html)

aws-cloudformation-stackset/aws-cloudformation-stackset.json
```diff
@@ -0,0 +1,309 @@
...
+        "TemplateURL": {
+            "description": "Location of file containing the template body. The URL must point to a template (max size: 460,800 bytes) that is located in an Amazon S3 bucket.",
+            "type": "string",
+            "minLength": 1,
+            "maxLength": 1024
+        }
+    },
+    "required": [
+        "PermissionModel"
+    ],
```
PatMyron, line 248:
```suggestion
```

["By default, SELF-MANAGED is specified."](https://docs.aws.amazon.com/cli/latest/reference/cloudformation/create-stack-set.html)

aws-cloudformation-stackset/aws-cloudformation-stackset.json
```diff
@@ -0,0 +1,309 @@
...
+    "additionalProperties": false,
+    "createOnlyProperties": [
+        "/properties/PermissionModel"
+    ],
+    "writeOnlyProperties": [
+        "/properties/TemplateURL",
+        "/properties/OperationPreferences"
+    ],
```
PatMyron, line 256:
```suggestion
```

Are `TemplateURL`s and `OperationPreferences` secrets or sensitive data?

aws-cloudformation-stackset/aws-cloudformation-stackset.json
```diff
@@ -0,0 +1,309 @@
...
+                "cloudformation:DescribeStackInstance"
+            ]
+        },
+        "update": {
+            "permissions": [
+                "s3:GetObject",
+                "cloudformation:UpdateStackSet",
+                "cloudformation:CreateStackInstances",
+                "cloudformation:DeleteStackInstances",
+                "cloudformation:UpdatetackInstances",
```
PatMyron, line 286:
```suggestion
                "cloudformation:UpdateStackInstances",
```

https://docs.aws.amazon.com/IAM/latest/UserGuide/list_awscloudformation.html#awscloudformation-actions-as-permissions

aws-cloudformation-stackset/aws-cloudformation-stackset.json
```diff
@@ -0,0 +1,309 @@
...
+                "cloudformation:TagResource",
+                "cloudformation:UntagResource"
+            ]
+        },
+        "delete": {
+            "permissions": [
+                "cloudformation:DeleteStackSet",
+                "cloudformation:DeleteStackInstances",
+                "cloudformation:DescribeStackSetOperation",
+                "cloudformation:TagResource"
```
PatMyron, line 297: Why does deleting a stack set require `cloudformation:TagResource`?

aws-cloudformation-stackset/aws-cloudformation-stackset.json
```diff
@@ -0,0 +1,309 @@
...
+        },
+        "Account": {
+            "description": "AWS account that you want to create stack instances in the specified Region(s) for.",
+            "type": "string",
+            "pattern": "^[0-9]{12}$"
+        },
+        "Region": {
+            "type": "string",
+            "pattern": "^[a-zA-Z0-9-]{1,128}$"
+        },
```
PatMyron, line 51: [Let's make sure we have these as common definitions and use those instead](https://github.com/aws-cloudformation/cloudformation-cli/issues/374)

aws-cloudformation-stackset/resource-role.yaml
```diff
@@ -0,0 +1,44 @@
...
+                - "cloudformation:DeleteStackSet"
+                - "cloudformation:DescribeStackInstance"
+                - "cloudformation:DescribeStackSet"
+                - "cloudformation:DescribeStackSetOperation"
+                - "cloudformation:ListStackInstances"
+                - "cloudformation:ListStackSets"
+                - "cloudformation:TagResource"
+                - "cloudformation:UntagResource"
+                - "cloudformation:UpdateStackSet"
+                - "cloudformation:UpdatetackInstances"
```
PatMyron, line 38:
```suggestion
                - "cloudformation:UpdateStackInstances"
```
https://docs.aws.amazon.com/IAM/latest/UserGuide/list_awscloudformation.html#awscloudformation-actions-as-permissions

aws-cloudformation-stackset/aws-cloudformation-stackset.json
xiwhuang: If user wants to change `PermissionModel`, they have to make sure that all stackinstances have been removed. By making `PermissionModel` as `createOnlyProperty` simplifies the implementation as framework would create a new stackset and replace which is pretty much what we trying to do here.

aws-cloudformation-stackset/aws-cloudformation-stackset.json
xiwhuang: I am wondering since `TemplateURL` and `OperationPreferences` can not be currently retrieved from service client, should we put them in `writeOnlyProperties`

aws-cloudformation-stackset/aws-cloudformation-stackset.json
xiwhuang: Nice catch. Will fix it.

aws-cloudformation-stackset/aws-cloudformation-stackset.json
xiwhuang: Nice catch. Will fix it.

aws-cloudformation-stackset/resource-role.yaml
xiwhuang: Nice catch. Will fix it.

aws-cloudformation-stackset/aws-cloudformation-stackset.json
```diff
@@ -0,0 +1,309 @@
...
+        "StackInstances": {
+            "description": "Stack instances in some specific accounts and Regions.",
+            "type": "object",
+            "properties": {
+                "DeploymentTargets": {
+                    "description": " The AWS OrganizationalUnitIds or Accounts for which to create stack instances in the specified Regions.",
+                    "type": "object",
+                    "properties": {
+                        "Accounts": {
+                            "description": "AWS accounts that you want to create stack instances in the specified Region(s) for.",
```
msaggar, line 87: can we say that we need a oneOf relationship here since we need either accounts or org ids?

https://json-schema.org/understanding-json-schema/reference/combining.html

aws-cloudformation-stackset/aws-cloudformation-stackset.json
msaggar: Nice idea! I think that we just need documentation about that on the resource

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/CallbackContext.java
```diff
@@ -0,0 +1,80 @@
...
+    private String operationId;
+
+    // Indicates initiation of analyzing template.
+    private boolean templateAnalyzed;
+
+    // Indicates initiation of resource stabilization.
+    private boolean stackSetCreated;
+
+    // Indicates initiation of stack instances creation.
+    private boolean addStacksStarted;
```
msaggar, line 31: this is confusing, for anything to track instances, can we include instance in the variable name instead of stacks because StackSets doesn't have a concept of `Stacks`

e.g: `private boolean addInstancesStarted`

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/CallbackContext.java
```diff
@@ -0,0 +1,80 @@
...
+
+    // Total running time
+    @Builder.Default
+    private int elapsedTime = 0;
+
+    private StackInstances stackInstancesInOperation;
+
+    // List to keep track on the complete status for creating
+    @Builder.Default
+    private Queue<StackInstances> createStacksQueue = new LinkedList<>();
```
msaggar, line 50: Same thing as mentioned above for these queues, StackSets works with instances not stacks

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/CallbackContext.java
xiwhuang: Renamed in new revision

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/CallbackContext.java
xiwhuang: Renamed in new revision

aws-cloudformation-stackset/aws-cloudformation-stackset.json
```diff
@@ -0,0 +1,311 @@
+{
```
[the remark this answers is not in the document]
xiwhuang: I am wondering CloudFormation might perform `drift-detection` on resources by leveraging Read Handler which means it might involves all properties we are defining here except `writeOnlyProperties`

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/CreateHandler.java
```diff
@@ -0,0 +1,96 @@
...
+import software.amazon.cloudformation.stackset.util.PhysicalIdGenerator;
+import software.amazon.cloudformation.stackset.util.Stabilizer;
+import software.amazon.cloudformation.stackset.util.Validator;
+
+import static software.amazon.cloudformation.stackset.translator.RequestTranslator.createStackSetRequest;
+import static software.amazon.cloudformation.stackset.util.Comparator.isAddingStackInstances;
+import static software.amazon.cloudformation.stackset.util.EnumUtils.Operations.ADD_INSTANCES;
+import static software.amazon.cloudformation.stackset.util.Stabilizer.getDelaySeconds;
+
+public class CreateHandler extends BaseHandler<CallbackContext> {
```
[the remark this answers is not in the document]
xiwhuang: When a `Create` fails and it needs to rollback, meaning it will delete the resource we trying to create here. Per the delete handler implementation, it is expected to detect the what instances to delete and remove the StackSet when StackInstances are stabilized to be deleted.
When a `Update` fails, it performs rolling back to previous version of template, meaning it kicks off an Update request which will be handled by Update Implementation here. StackInstances will be updated/deleted/created per circumstances, as well as the other configurations of StackSet.

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/util/OperationOperator.java
```diff
@@ -0,0 +1,160 @@
...
+        context.setStackInstancesInOperation(instancesQueue.remove());
+        return response.operationId();
+    }
+
+    /**
+     * Invokes DeleteStackInstances API to delete old {@link StackInstances} based on {@link CallbackContext#getDeleteStacksQueue()}
+     * @return Operation Id from {@link DeleteStackInstancesResponse}
+     */
+    private String deleteStackInstances() {
+        final Queue<StackInstances> instancesQueue = context.getDeleteStacksQueue();
```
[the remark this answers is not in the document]
xiwhuang: If ProgressEvent returns `IN_PROGRESS` with `CallbackContext` which means resource needs to stabilize. CallbackContext containing all necessary states ensures that lambda or whatever compute engine it can be picking up where last invocation stands.

aws-cloudformation-stackset/aws-cloudformation-stackset.json
xiwhuang: Updated to using common definitions

aws-cloudformation-stackset/aws-cloudformation-stackset.json
xiwhuang: Tested with latest version of plug-in, still facing validation error using `oneOf` in this case. This will be validated inside handler anyway.

aws-cloudformation-stackset/aws-cloudformation-stackset.json
```diff
@@ -0,0 +1,311 @@
...
+    "typeName": "AWS::CloudFormation::StackSet",
+    "description": "Resource Type definition for AWS::CloudFormation::StackSet",
+    "sourceUrl": "https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-cloudformation.git",
```
PatMyron, line 4:
```suggestion
    "sourceUrl": "https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-cloudformation.git",
    "resourceLink": {
        "templateUri": "/cloudformation/home?region=${awsRegion}#/stacksets/${StackSetId}",
        "mappings": {
            "StackSetId": "/StackSetId"
        }
    },
```

https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/pull/67/

xiwhuang: That's definitely good to have here. Will add in next commit.

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/BaseHandlerStd.java
```diff
@@ -0,0 +1,247 @@
...
+            final ResourceHandlerRequest<ResourceModel> request,
+            final CallbackContext callbackContext,
+            final ProxyClient<CloudFormationClient> proxyClient,
+            final Logger logger);
+
+    protected boolean filterException(AwsRequest request, Exception e, ProxyClient<CloudFormationClient> client, ResourceModel model, CallbackContext context) {
+        return e instanceof OperationInProgressException;
+    }
+
+    protected ProgressEvent<ResourceModel, CallbackContext> createStackInstances(
```
ammokhov, line 110: can u move it to `CreateHandler` if it's used only there?

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/BaseHandlerStd.java
```diff
@@ -0,0 +1,247 @@
...
+                    })
+                    .stabilize((request, response, proxyInvocation, resourceModel, context) -> isOperationStabilized(proxyInvocation, resourceModel, response.operationId(), logger))
+                    .handleError((request, e, proxyClient, model_, context) -> {
+                        // If StackInstanceNotFoundException is thrown by the service, then we did succeed delete/stabilization call in case of out of band deletion.
+                        if (e instanceof StackInstanceNotFoundException) {
+                            return ProgressEvent.success(model_, context);
+                        }
+                        if (e instanceof OperationInProgressException) {
+                            return ProgressEvent.progress(model_, context);
+                        }
```
ammokhov, line 169: should this go to a retry filter function?

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/BaseHandlerStd.java
```diff
@@ -0,0 +1,247 @@
...
+
+            if (!progressEvent.isSuccess()) {
+                return progressEvent;
+            }
+        }
+
+        return ProgressEvent.progress(model, callbackContext);
+    }
+
+    protected ProgressEvent<ResourceModel, CallbackContext> deleteStackInstances(
```
ammokhov, line 141: ditto: move it to `DeleteHandler`

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/DeleteHandler.java
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * Implement client invocation of the delete request through the proxyClient, which is already initialised with
+     * caller credentials, correct region and retry settings
+     *
+     * @param proxy    Amazon webservice proxy to inject credentials correctly.
+     * @param client   the aws service client to make the call
+     * @param progress event of the previous state indicating success, in progress with delay callback or failed state
+     * @return delete resource response
+     */
+    protected ProgressEvent<ResourceModel, CallbackContext> deleteStackSet(
```
ammokhov, line 47: can be private right?

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/ListHandler.java
```diff
@@ -0,0 +1,45 @@
...
+                .stream()
+                .map(stackSetSummary -> ResourceModelBuilder.builder()
+                        .proxyClient(proxyClient)
+                        .stackSet(describeStackSet(proxyClient, stackSetSummary.stackSetId()))
+                        .build().buildModel())
+                .collect(Collectors.toList());
+
+        return ProgressEvent.<ResourceModel, CallbackContext>builder()
+                .resourceModels(models)
+                .status(OperationStatus.SUCCESS)
```
ammokhov, line 42: no `nextToken`?

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/ReadHandler.java
```diff
@@ -0,0 +1,34 @@
...
+            final CallbackContext callbackContext,
+            final ProxyClient<CloudFormationClient> proxyClient,
+            final Logger logger) {
+
+        this.logger = logger;
+        final ResourceModel model = request.getDesiredResourceState();
+
+        return ProgressEvent.<ResourceModel, CallbackContext>builder()
+                .resourceModel(ResourceModelBuilder.builder()
+                        .proxyClient(proxyClient)
```
ammokhov, line 28: why do you need to return proxyClient?

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/ListHandler.java
```diff
@@ -0,0 +1,45 @@
...
+            final Logger logger) {
+
+        final ListStackSetsResponse response = proxyClient.injectCredentialsAndInvokeV2(
+                listStackSetsRequest(request.getNextToken()), proxyClient.client()::listStackSets);
+
+        final List<ResourceModel> models = response
+                .summaries()
+                .stream()
+                .map(stackSetSummary -> ResourceModelBuilder.builder()
+                        .proxyClient(proxyClient)
```
ammokhov, line 35: rm

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/ReadHandler.java
```diff
@@ -0,0 +1,34 @@
...
+            final AmazonWebServicesClientProxy proxy,
+            final ResourceHandlerRequest<ResourceModel> request,
+            final CallbackContext callbackContext,
+            final ProxyClient<CloudFormationClient> proxyClient,
+            final Logger logger) {
+
+        this.logger = logger;
+        final ResourceModel model = request.getDesiredResourceState();
+
+        return ProgressEvent.<ResourceModel, CallbackContext>builder()
```
ammokhov, line 26: plz use 
```java
return proxy.initiate()
    .translate(...)
    .makeServiceCall(...)
    ...
```

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/UpdateHandler.java
```diff
@@ -0,0 +1,95 @@
...
+        final ResourceModel model = request.getDesiredResourceState();
+        final ResourceModel previousModel = request.getPreviousResourceState();
+        final StackInstancesPlaceHolder placeHolder = new StackInstancesPlaceHolder();
+        analyzeTemplate(proxy, previousModel, placeHolder, model);
+
+        return ProgressEvent.progress(model, callbackContext)
+                .then(progress -> updateStackSet(proxy, proxyClient, progress, previousModel))
+                .then(progress -> deleteStackInstances(proxy, proxyClient, progress, placeHolder.getDeleteStackInstances(), logger))
+                .then(progress -> createStackInstances(proxy, proxyClient, progress, placeHolder.getCreateStackInstances(), logger))
+                .then(progress -> updateStackInstances(proxy, proxyClient, progress, placeHolder.getUpdateStackInstances(), logger))
```
ammokhov, line 39: why are we doing it - looks like an immutable update? looks like such kinda thing should be handled outside of update handler @PatMyron

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/UpdateHandler.java
```diff
@@ -0,0 +1,95 @@
...
+        final ResourceModel desiredModel = progress.getResourceModel();
+        final CallbackContext callbackContext = progress.getCallbackContext();
+
+        if (isStackSetConfigEquals(previousModel, desiredModel)) {
+            return ProgressEvent.progress(desiredModel, callbackContext);
+        }
+        return proxy.initiate("AWS-CloudFormation-StackSet::UpdateStackSet", client, desiredModel, callbackContext)
+                .translateToServiceRequest(modelRequest -> updateStackSetRequest(modelRequest))
+                .makeServiceCall((modelRequest, proxyInvocation) -> {
+                    UpdateStackSetResponse response = proxyInvocation.injectCredentialsAndInvokeV2(modelRequest, proxyInvocation.client()::updateStackSet);
```
ammokhov, line 68: `final`?

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/UpdateHandler.java
```diff
@@ -0,0 +1,95 @@
...
+    private void analyzeTemplate(
+            final AmazonWebServicesClientProxy proxy,
+            final ResourceModel previousModel,
+            final StackInstancesPlaceHolder placeHolder,
+            final ResourceModel model) {
+
+        new Validator().validateTemplate(proxy, model.getTemplateBody(), model.getTemplateURL(), logger);
+        InstancesAnalyzer.builder().desiredModel(model).previousModel(previousModel).build()
+                .analyzeForUpdate(placeHolder);
+    }
```
ammokhov, line 94: a good candidate to be moved to `BaseHandlerStd` and reused by all handlers if needed

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/util/AwsCredentialsExtractor.java
```diff
@@ -0,0 +1,248 @@
...
+import java.util.List;
+import java.util.Objects;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+/**
+ * Utility class to extract AWS Credentials Provider from {@link AmazonWebServicesClientProxy}.
+ * <p>
+ * Because {@link AmazonWebServicesClientProxy#injectCredentialsAndInvokeV2(AwsRequest, Function)} doesn't extend
+ * {@link ResponseInputStream<AwsResponse>}, but S3 GetObject requires AWS Credentials Provider to authenticate user,
```
jotompki, line 25: if theres a miss in the proxy feel free to add a method to class to support it: https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/blob/master/src/main/java/software/amazon/cloudformation/proxy/AmazonWebServicesClientProxy.java#L58 so more customers can take advantage rather than add in this extractor

xiwhuang: Sounds good, will work on adding a new method for that.

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/BaseHandlerStd.java
xiwhuang: `UpdateHandler` will be using this as well

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/DeleteHandler.java
xiwhuang: Sure, we can scope it down to private.

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/UpdateHandler.java
xiwhuang: Sure.

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/UpdateHandler.java
xiwhuang: Sure

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/UpdateHandler.java
xiwhuang: I am not sure this part, are you saying `updateStackInstances` API is an immutable update?

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/ReadHandler.java
xiwhuang: Actually I feel like using a custom builder to construct resource models will be preferable here as it can also be reusable in ListHandler.

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/BaseHandlerStd.java
xiwhuang: it will be reused in update handler as well.

jotompki:
any way to scope this down a little more into separate PRs?

Having one for the auto generated code, then the schema, then maybe individual handlers will let us as reviewers more readily give feedback

ammokhov:
> any way to scope this down a little more into separate PRs?
> 
> Having one for the auto generated code, then the schema, then maybe individual handlers will let us as reviewers more readily give feedback

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/ListHandler.java
xiwhuang: nice catch, will add.

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/ReadHandler.java
xiwhuang: That is a custom resource model builder. This would only return a StackSet ResourceModel

xiwhuang:
> any way to scope this down a little more into separate PRs?
> 
> Having one for the auto generated code, then the schema, then maybe individual handlers will let us as reviewers more readily give feedback

It is not easy to start it over though, But I think it is a very good suggestion, maybe we should make it as a documented PR standard and let everyone follows.

aws-cloudformation-stackset/pom.xml
```diff
@@ -0,0 +1,251 @@
...
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>s3</artifactId>
+        </dependency>
+        <!-- https://mvnrepository.com/artifact/com.amazonaws/aws-java-sdk-s3 -->
+        <dependency>
+            <groupId>com.amazonaws</groupId>
+            <artifactId>aws-java-sdk-s3</artifactId>
+            <version>1.11.759</version>
+        </dependency>
```
PatMyron, line 67: what's the reasoning for both of these S3 SDK dependencies?

aws-cloudformation-stackset/pom.xml
```diff
@@ -0,0 +1,251 @@
...
+        <dependency>
+            <groupId>org.yaml</groupId>
+            <artifactId>snakeyaml</artifactId>
+            <version>1.26</version>
+        </dependency>
+        <!-- https://mvnrepository.com/artifact/software.amazon.awssdk/s3 -->
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>s3</artifactId>
+        </dependency>
```
PatMyron, line 61: might be helpful to pin versions for all dependencies as well

xiwhuang: The version is pined in `dependencyManagement ` which is `<version>2.10.70</version>`

aws-cloudformation-stackset/pom.xml
xiwhuang: V1 S3 is for constructing and validating an S3URL. V2 is for invoking `s3::GetObject` API

aws-cloudformation-stackset/aws-cloudformation-stackset.json
PatMyron: @xiwhuang might need to bump this:
https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-cloudformation/blob/1e7ca3fc2d9223b7ac244ca08258991b4216e2b8/aws-cloudformation-stackset/pom.xml#L40-L45

https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/compare/v0.1.4...master

https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/compare/v2.0.2...master

https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/pull/248/files

PatMyron: actually upgrading [`cloudformation-cli`](https://github.com/aws-cloudformation/cloudformation-cli/) with `pip3 install cloudformation-cli --upgrade` solves this instead

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/util/AwsCredentialsExtractor.java
xiwhuang: Have submitted a [PR](https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/pull/259) to add those methods

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/ListHandler.java
xiwhuang: Again that's a parameter to construct `ResourceModelBuilder` for `buildModel()`

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/ReadHandler.java
ammokhov: `ListHandler` requires to return only primary identifiers so it is a bit off the context. 
`ReadHandler` is used to get consistent results for `CreateHandler` and `UpdateHandler`. CallChain delegates finally to read handler to ensure complete state is read and sent back after setting up identifiers and other information.

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/ReadHandler.java
ammokhov: could you give a bit more context?

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/BaseHandlerStd.java
ammokhov: actually, it's fine to keep it here

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/ReadHandler.java
xiwhuang: As discussed offline, [ResourceModelBuilder](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-cloudformation/pull/6/files/1e7ca3fc2d9223b7ac244ca08258991b4216e2b8#diff-7da75b18ea93d386e862770acd11b5dd) will leverage `ProxyClient` to invoke other API services and retrieve all the attributes that are required to construct a `ResourceModel` that returning to customers. And of course, proxyClient will not be included in the Model.

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/util/Validator.java
```diff
@@ -0,0 +1,176 @@
...
+            final Logger logger) {
+
+        if (Strings.isNullOrEmpty(templateBody) == Strings.isNullOrEmpty(templateLocation)) {
+            throw new CfnInvalidRequestException("Exactly one of TemplateBody or TemplateUrl must be specified");
+        }
+        String content = null;
+        try {
+            if (!Strings.isNullOrEmpty(templateLocation)) {
+                validateS3Uri(templateLocation);
+                content = getUrlContent(proxy, templateLocation);
```
jotompki, line 147: I would do a size check with getObjectMetadata before copying over to ensure we arent trying to copy over a crazy huge template

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/util/Comparator.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Compares if desired model uses the same stack set configs other than stack instances
+     * when it comes to updating the resource
+     *
+     * @param previousModel previous {@link ResourceModel}
+     * @param desiredModel  desired {@link ResourceModel}
+     * @return
+     */
+    public static boolean isStackSetConfigEquals(
```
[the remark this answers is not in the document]
xiwhuang: Yup, we only compare a subset of properties, for property like `OperationPerferences` which is one-time use property we don't need to care about it. And if `TemplateURL` is specified, we have to call the UpdateAPI whatever.

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/util/Comparator.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    /**
+     * Compares if two collections equal in a null-safe way.
+     *
+     * @param collection1
+     * @param collection2
+     * @return boolean indicates if two collections equal.
+     */
+    public static boolean isEquals(final Collection<?> collection1, final Collection<?> collection2) {
```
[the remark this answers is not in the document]
xiwhuang: Sure.

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/util/Validator.java
xiwhuang: Added a size check for that

aws-cloudformation-stackset/aws-cloudformation-stackset.json
```diff
@@ -0,0 +1,320 @@
...
+                "ParameterKey": {
+                    "description": "The key associated with the parameter. If you don't specify a key and value for a particular parameter, AWS CloudFormation uses the default value that is specified in your template.",
+                    "type": "string"
+                },
+                "ParameterValue": {
+                    "description": "The input value associated with the parameter.",
+                    "type": "string"
+                }
+            },
+            "additionalProperties": false
```
jotompki, line 84: should both key/value be required here? not requiring could lead to some npes if customer specifies an empty parameter object

aws-cloudformation-stackset/aws-cloudformation-stackset.json
jotompki: have you retested this with new version of plugin? what is the result

aws-cloudformation-stackset/aws-cloudformation-stackset.json
```diff
@@ -0,0 +1,320 @@
...
+                            "insertionOrder": false,
+                            "items": {
+                                "$ref": "#/definitions/OrganizationalUnitId"
+                            }
+                        }
+                    }
+                },
+                "Regions": {
+                    "description": "The names of one or more Regions where you want to create stack instances using the specified AWS account(s).",
+                    "type": "array",
```
jotompki, line 116: "one or more" - use [minItems keyword](https://json-schema.org/understanding-json-schema/reference/array.html) to enforce this

aws-cloudformation-stackset/aws-cloudformation-stackset.json
```diff
@@ -0,0 +1,320 @@
...
+                    }
+                }
+            },
+            "required": [
+                "DeploymentTargets",
+                "Regions"
+            ]
+        }
+    },
+    "properties": {
```
jotompki, line 139: nit: i think for consistency we should strive to normalize any nested object schemas in the definitions section. not sure why some are normalized and defined there but others like OperationPreferences and Deployment targets are nested.

aws-cloudformation-stackset/aws-cloudformation-stackset.json
```diff
@@ -0,0 +1,320 @@
...
+            "maxLength": 51200
+        },
+        "TemplateURL": {
+            "description": "Location of file containing the template body. The URL must point to a template (max size: 460,800 bytes) that is located in an Amazon S3 bucket.",
+            "type": "string",
+            "minLength": 1,
+            "maxLength": 1024
+        }
+    },
+    "required": [
```
jotompki, line 254: isn't templateBody or templateURL required? the only thing necessary to create a stackset is the permissions model?

aws-cloudformation-stackset/aws-cloudformation-stackset.json
```diff
@@ -0,0 +1,320 @@
...
+    "additionalProperties": false,
+    "createOnlyProperties": [
+        "/properties/PermissionModel"
+    ],
+    "writeOnlyProperties": [
+        "/properties/TemplateURL",
+        "/properties/OperationPreferences"
+    ],
+    "readOnlyProperties": [
+        "/properties/StackSetId"
```
jotompki, line 266: is this an ARN? i feel like it will be useful to customers to return a stacksets arn

aws-cloudformation-stackset/aws-cloudformation-stackset.json
jotompki, line 139: A customer can't specify a stackset name?

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/BaseHandlerStd.java
```diff
@@ -0,0 +1,317 @@
...
+            final ProgressEvent<ResourceModel, CallbackContext> progressEvent = proxy
+                    .initiate("AWS-CloudFormation-StackSet::CreateStackInstances" + stackInstances.hashCode(), client, model, callbackContext)
+                    .translateToServiceRequest(modelRequest -> createStackInstancesRequest(modelRequest.getStackSetId(), modelRequest.getOperationPreferences(), stackInstances))
+                    .backoffDelay(MULTIPLE_OF)
+                    .makeServiceCall((modelRequest, proxyInvocation) -> {
+                        final CreateStackInstancesResponse response = proxyInvocation.injectCredentialsAndInvokeV2(modelRequest, proxyInvocation.client()::createStackInstances);
+                        logger.log(String.format("%s CreateStackInstances in [%s] of [%s] initiated", ResourceModel.TYPE_NAME, stackInstances.getRegions(), stackInstances.getDeploymentTargets()));
+                        return response;
+                    })
+                    .stabilize((request, response, proxyInvocation, resourceModel, context) -> isOperationStabilized(proxyInvocation, resourceModel, response.operationId(), logger))
```
jotompki, line 144: does the proxy work for looping for longer running stack creations? im a bit uneasy having this looping over these long running operations because of timeouts. have you tested this with a sufficient number of stack instances that all are a bit longer running?

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/util/TemplateParser.java
```diff
@@ -0,0 +1,173 @@
...
+     * Deserializes YAML from template content string
+     *
+     * @param templateString Template content
+     * @return Template map
+     * @throws ParseException if fails to parse the template
+     */
+    @VisibleForTesting
+    protected static Map<String, Object> deserializeYaml(final String templateString) {
+        try {
+            final Map<String, Object> template = new Yaml().load(templateString);
```
jotompki, line 93: does this handle our shorthand abbreviations like !Ref and !GetAtt https://aws.amazon.com/blogs/aws/aws-cloudformation-update-yaml-cross-stack-references-simplified-substitution/

Also to save some code: [yaml is a superset of json](https://yaml.org/spec/1.2/spec.html#id2759572) so you dont need  the json parser

aws-cloudformation-stackset/aws-cloudformation-stackset.json
xiwhuang: Nice catch!

aws-cloudformation-stackset/aws-cloudformation-stackset.json
xiwhuang: The logical ID is the stackset name

aws-cloudformation-stackset/aws-cloudformation-stackset.json
xiwhuang: updated in the latest commit.

aws-cloudformation-stackset/aws-cloudformation-stackset.json
xiwhuang: updated in the latest commit.

aws-cloudformation-stackset/aws-cloudformation-stackset.json
xiwhuang: resolved as discussed offline

aws-cloudformation-stackset/aws-cloudformation-stackset.json
xiwhuang: updated in the latest commit.

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/BaseHandlerStd.java
xiwhuang: resolved as discussed offline,

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/util/TemplateParser.java
xiwhuang: updated in the latest commit.

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/util/Validator.java
```diff
@@ -0,0 +1,189 @@
...
+import static software.amazon.cloudformation.stackset.translator.RequestTranslator.getObjectRequest;
+import static software.amazon.cloudformation.stackset.translator.RequestTranslator.headObjectRequest;
+import static software.amazon.cloudformation.stackset.util.TemplateParser.deserializeTemplate;
+import static software.amazon.cloudformation.stackset.util.TemplateParser.getMapFromTemplate;
+import static software.amazon.cloudformation.stackset.util.TemplateParser.getStringFromTemplate;
+
+/**
+ * Utility class to validate properties in {@link software.amazon.cloudformation.stackset.ResourceModel}
+ */
+public class Validator {
```
PatMyron, line 25: this validation needs to be done service-side

aws-cloudformation-stackset/aws-cloudformation-stackset.json
PatMyron: don't think logical IDs should be used this way

this would force a more dangerous [`replacementStrategy`](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/pull/86) when [`createOnlyProperties`](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/) are modified in order to avoid [`NameAlreadyExists`](https://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_CreateStackSet.html#API_CreateStackSet_Errors)

msaggar: Good call out Pat. This is not the only case where we have to consider the replacement strategy. For example, consider a Create,Delete replacement strategy. You have StackSet A with 10 instances, something triggers a replacement strategy and you now CreateStackSet A again (this would fail for the reason pat is calling out) but even if it did not you would could hit a case where you are deploying the same resources in an instance to a set of target accounts. If Account 1 had 2 queues defined in the template and then create was called again with those same queue resources then stack creation would fail because the resource already exists. Extend that possibility to the N instances that belong to a stackset. If the threshold for DeploymentPreferences is breached then the deployment would fail and you would be stuck.

We will have to use a DeleteCreate strategy to get around both of these, but I am in favor of making the name a create only property of the resource.

xiwhuang: Agreed. We should make `StackSetName` as required and a `CreateOnlyProperty`

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/util/Validator.java
xiwhuang: Have updated to use API `GetTemplateSummary` to validate template.

aws-cloudformation-stackset/aws-cloudformation-stackset.json
```diff
@@ -0,0 +1,354 @@
+{
+    "typeName": "AWS::CloudFormation::StackSet",
+    "description": "Resource Type definition for AWS::CloudFormation::StackSet",
```
miparnisari, line 3: This is the description that is shown in the Console UI. Please write something more meaningful here.

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/DeleteHandler.java
```diff
@@ -0,0 +1,65 @@
...
+
+        final ResourceModel model = progress.getResourceModel();
+        final CallbackContext callbackContext = progress.getCallbackContext();
+
+        return proxy.initiate("AWS-CloudFormation-StackSet::DeleteStackSet", client, model, callbackContext)
+                .translateToServiceRequest(modelRequest -> deleteStackSetRequest(modelRequest.getStackSetId()))
+                .makeServiceCall((modelRequest, proxyInvocation) -> {
+                    final DeleteStackSetResponse response = proxyInvocation.injectCredentialsAndInvokeV2(
+                            deleteStackSetRequest(model.getStackSetId()), proxyInvocation.client()::deleteStackSet);
+                    logger.log(String.format("%s successfully deleted.", ResourceModel.TYPE_NAME));
```
miparnisari, line 60:
```suggestion
                    logger.log(String.format("%s successfully deleted.", model.getStackSetId())));
```

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/translator/PropertyTranslator.java
```diff
@@ -0,0 +1,172 @@
...
+        return parameters.stream()
+                .map(parameter -> Parameter.builder()
+                        .parameterKey(parameter.getParameterKey())
+                        .parameterValue(parameter.getParameterValue())
+                        .build())
+                .collect(Collectors.toList());
+    }
+
+    /**
+     * Converts resource model Parameters to StackSet SDK Parameters
```
miparnisari, line 84:
```suggestion
     * Converts StackSet SDK Parameters to resource model Parameters
```

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/BaseHandlerStd.java
```diff
@@ -0,0 +1,315 @@
...
+     * @param operationId Operation ID
+     * @return boolean
+     */
+    @VisibleForTesting
+    protected static boolean isStackSetOperationDone(
+            final StackSetOperationStatus status, final String operationId, final Logger logger) {
+
+        switch (status) {
+            case SUCCEEDED:
+                logger.log(String.format("%s has been successfully stabilized.", operationId));
```
miparnisari, line 79:
```suggestion
                logger.log(String.format("StackSet operation ID %s: successfully stabilized.", operationId));
```

aws-cloudformation-stackset/aws-cloudformation-stackset.json
xiwhuang: Updated a short description for StackSet resource.

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/DeleteHandler.java
xiwhuang: updated

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/translator/PropertyTranslator.java
xiwhuang: Fixed, nice catch

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/BaseHandlerStd.java
xiwhuang: Updated.

guanghug:
Reviewed the StackSets logics are fine.

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/util/Validator.java
```diff
@@ -0,0 +1,55 @@
...
+     * @param templateLocation {@link software.amazon.cloudformation.stackset.ResourceModel#getTemplateURL}
+     * @throws CfnInvalidRequestException if template is not valid
+     */
+    public void validateTemplate(
+            final ProxyClient<CloudFormationClient> proxyClient,
+            final String templateBody,
+            final String templateLocation) {
+
+        final GetTemplateSummaryResponse response = proxyClient.injectCredentialsAndInvokeV2(
+                getTemplateSummaryRequest(templateBody, templateLocation),
```
wbingli, line 48: NIce!

aws-cloudformation-stackset/aws-cloudformation-stackset.json
```diff
@@ -0,0 +1,301 @@
...
+        "Region": {
+            "type": "string",
+            "pattern": "^[a-zA-Z0-9-]{1,128}$"
+        },
+        "OrganizationalUnitId": {
+            "type": "string",
+            "pattern": "^(ou-[a-z0-9]{4,32}-[a-z0-9]{8,32}|r-[a-z0-9]{4,32})$"
+        },
+        "Capability": {
+            "type": "string",
```
yinglonw, line 65: This should be an array of the capabilities.

xiwhuang: Yes, capabilities should be an array. This is the `Capability` in `definitions` which will be referenced in [`Capabilities` in `properties`](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-cloudformation/pull/6/files/705bb41f86bd159f2950c39cff3bcd999cb57f88#diff-1d6fda833e3f951be4531e864f3d030cR146).

s0enke:
Great work! Is there a timeframe for bringing this to official CloudFormation (e.g. https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/ReleaseHistory.html)?

ikben:
It might also be useful to change the state in the coverage roadmap:
https://github.com/aws-cloudformation/aws-cloudformation-coverage-roadmap/issues/102


Additionally is their work planned for having separate StackInstance resources? I can see value in being able to define the StackSet in one Stack and StackInstances in another (https://github.com/aws-cloudformation/aws-cloudformation-coverage-roadmap/issues/103).

PR #11, opened 2020-08-21
Removing unecessary permission

rjlohan:
*Issue #, if available:* N/A

*Description of changes:* Schema included a permission which is used by the downstream service, rather than the handler itself and so isn't necessary here.

### DIFF
--- a/aws-cloudformation-resourceversion/aws-cloudformation-resourceversion.json
+++ b/aws-cloudformation-resourceversion/aws-cloudformation-resourceversion.json
@@ -131,8 +131,7 @@
       "permissions": [
         "cloudformation:DescribeTypeRegistration",
         "cloudformation:ListTypeVersions",
-        "cloudformation:RegisterType",
-        "iam:PassRole"
+        "cloudformation:RegisterType"
       ]
     },
     "read": {


aws-cloudformation-resourceversion/aws-cloudformation-resourceversion.json
```diff
@@ -131,8 +131,7 @@
       "permissions": [
         "cloudformation:DescribeTypeRegistration",
         "cloudformation:ListTypeVersions",
-        "cloudformation:RegisterType",
-        "iam:PassRole"
```
wbingli, line 135: If register-type  with execution role, the `iam:PassRole` will actually be required.

rjlohan: But only indirectly by the `RegisterType` API. Should it still be shown here? Seems a gray area for me if the handler itself isn't invoking the API listed here...

rjlohan: No you're right, it should be declared here as it's part of the auth check when calling `RegisterType`.

PR #19, opened 2020-10-02
ResourceDefaultVersion - added Primary ID TypeArn and made delete han…

ugudip:
1. Added Primary ID for ResourceDefaultVersion - TypeArn
2. Resolved few bugs - version conflicts and exceptions
3. Made unit test changes, jacoco coverage >80%

### DIFF
--- a/aws-cloudformation-resourcedefaultversion/aws-cloudformation-resourcedefaultversion.json
+++ b/aws-cloudformation-resourcedefaultversion/aws-cloudformation-resourcedefaultversion.json
@@ -3,7 +3,7 @@
   "description": "The default version of a resource that has been registered in the CloudFormation Registry.",
   "sourceUrl": "https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-cloudformation",
   "properties": {
-    "Arn": {
+    "TypeVersionArn": {
       "description": "The Amazon Resource Name (ARN) of the type version.",
       "pattern": "^arn:aws[A-Za-z0-9-]{0,64}:cloudformation:[A-Za-z0-9-]{1,64}:([0-9]{12})?:type/resource/.+$",
       "type": "string"
@@ -13,8 +13,8 @@
       "pattern": "^[A-Za-z0-9]{2,64}::[A-Za-z0-9]{2,64}::[A-Za-z0-9]{2,64}$",
       "type": "string"
     },
-    "TypeArn": {
-      "description": "The Amazon Resource Name (ARN) of the type.",
+    "Arn": {
+      "description": "The Amazon Resource Name (ARN) of the type. This is used to uniquely identify a ResourceDefaultVersion",
       "pattern": "^arn:aws[A-Za-z0-9-]{0,64}:cloudformation:[A-Za-z0-9-]{1,64}:([0-9]{12})?:type/resource/.+$",
       "type": "string"
     },
@@ -27,7 +27,7 @@
   "oneOf": [
     {
       "required": [
-        "Arn"
+        "TypeVersionArn"
       ]
     },
     {
@@ -37,8 +37,11 @@
       ]
     }
   ],
+  "readOnlyProperties": [
+    "/properties/Arn"
+  ],
   "primaryIdentifier": [
-    "/properties/TypeArn"
+    "/properties/Arn"
   ],
   "handlers": {
     "create": {
--- a/aws-cloudformation-resourcedefaultversion/aws-cloudformation-resourcedefaultversion.json
+++ b/aws-cloudformation-resourcedefaultversion/aws-cloudformation-resourcedefaultversion.json
@@ -13,6 +13,11 @@
       "pattern": "^[A-Za-z0-9]{2,64}::[A-Za-z0-9]{2,64}::[A-Za-z0-9]{2,64}$",
       "type": "string"
     },
+    "TypeArn": {
+      "description": "The Amazon Resource Name (ARN) of the type.",
+      "pattern": "^arn:aws[A-Za-z0-9-]{0,64}:cloudformation:[A-Za-z0-9-]{1,64}:([0-9]{12})?:type/resource/.+$",
+      "type": "string"
+    },
     "VersionId": {
       "description": "The ID of an existing version of the resource to set as the default.",
       "pattern": "^[A-Za-z0-9-]{1,128}$",
@@ -33,7 +38,7 @@
     }
   ],
   "primaryIdentifier": [
-    "/properties/Arn"
+    "/properties/TypeArn"
   ],
   "handlers": {
     "create": {
--- a/aws-cloudformation-resourcedefaultversion/src/main/java/software/amazon/cloudformation/resourcedefaultversion/CreateHandler.java
+++ b/aws-cloudformation-resourcedefaultversion/src/main/java/software/amazon/cloudformation/resourcedefaultversion/CreateHandler.java
@@ -12,23 +12,22 @@ public class CreateHandler extends BaseHandlerStd {
 
     @Override
     public ProgressEvent<ResourceModel, CallbackContext> handleRequest(
-        final AmazonWebServicesClientProxy proxy,
-        final ResourceHandlerRequest<ResourceModel> request,
-        final CallbackContext callbackContext,
-        final ProxyClient<CloudFormationClient> proxyClient,
-        final Logger logger) {
+            final AmazonWebServicesClientProxy proxy,
+            final ResourceHandlerRequest<ResourceModel> request,
+            final CallbackContext callbackContext,
+            final ProxyClient<CloudFormationClient> proxyClient,
+            final Logger logger) {
         final ResourceModel resourceModel = request.getDesiredResourceState();
-        if(resourceModel.getTypeArn() == null)
-        {
-          String generatedTypeArn = createTypeArn(request);
-          resourceModel.setTypeArn(generatedTypeArn);
-          return ProgressEvent.progress(resourceModel, callbackContext);
+        if (StringUtils.isNullOrEmpty(resourceModel.getArn())) {
+            String generatedArn = createArn(request);
+            resourceModel.setArn(generatedArn);
+            return ProgressEvent.progress(resourceModel, callbackContext);
         }
         return ProgressEvent.progress(resourceModel, callbackContext)
                 .then(progress -> {
                     final ResourceModel model = progress.getResourceModel();
                     logger.log(String.format("Creating [Arn: %s | Type: %s | Version: %s]",
-                            model.getArn(), model.getTypeName(), model.getVersionId()));
+                            model.getTypeVersionArn(), model.getTypeName(), model.getVersionId()));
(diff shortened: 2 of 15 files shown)


aws-cloudformation-resourcedefaultversion/aws-cloudformation-resourcedefaultversion.json
```diff
@@ -13,6 +13,11 @@
       "pattern": "^[A-Za-z0-9]{2,64}::[A-Za-z0-9]{2,64}::[A-Za-z0-9]{2,64}$",
       "type": "string"
     },
+    "TypeArn": {
```
tyangjie, line 16: rename "Arn" to "TypeVersionArn"?

tyangjie, line 16: Rename "Arn" to "TypeVersionArn"?

aws-cloudformation-resourcedefaultversion/src/main/java/software/amazon/cloudformation/resourcedefaultversion/CreateHandler.java
```diff
@@ -23,37 +17,40 @@
...
-
-        return initiator
-            .translateToServiceRequest(Translator::translateToUpdateRequest) // same API call for Create + Update
-            .makeServiceCall((r, c) -> createResource(resourceModel, r, c))
-            .done(setTypeDefaultVersionResponse -> new ReadHandler().handleRequest(proxy, request, callbackContext, proxyClient, logger));
+        if(resourceModel.getTypeArn() == null)
+        {
+          String generatedTypeArn = createTypeArn(request);
+          resourceModel.setTypeArn(generatedTypeArn);
+          return ProgressEvent.progress(resourceModel, callbackContext);
```
tyangjie, line 25: if type arn is null from the template input, then create handler does nothing?

aws-cloudformation-resourcedefaultversion/src/main/java/software/amazon/cloudformation/resourcedefaultversion/CreateHandler.java
```diff
@@ -23,37 +17,40 @@
...
-
-        logger.log(String.format("Creating [Arn: %s | Type: %s | Version: %s]",
-            resourceModel.getArn(), resourceModel.getTypeName(), resourceModel.getVersionId()));
-
-        return initiator
-            .translateToServiceRequest(Translator::translateToUpdateRequest) // same API call for Create + Update
-            .makeServiceCall((r, c) -> createResource(resourceModel, r, c))
-            .done(setTypeDefaultVersionResponse -> new ReadHandler().handleRequest(proxy, request, callbackContext, proxyClient, logger));
+        if(resourceModel.getTypeArn() == null)
+        {
```
tyangjie, line 22: Use ```StringUtils.isNotEmpty```,  and please take care of spacing/formatting

aws-cloudformation-resourcedefaultversion/src/main/java/software/amazon/cloudformation/resourcedefaultversion/CreateHandler.java
```diff
@@ -23,37 +17,40 @@
...
-        final SetTypeDefaultVersionRequest awsRequest,
-        final ProxyClient<CloudFormationClient> proxyClient) {
-
-        boolean exists = exists(proxyClient, model);
-        if (exists) {
-            this.logger.log("Default version for type is already set as requested");
-            throw new CfnAlreadyExistsException(ResourceModel.TYPE_NAME, model.getPrimaryIdentifier().toString());
+    private String createTypeArn(ResourceHandlerRequest<ResourceModel> request) {
+        final ResourceModel resourceModel = request.getDesiredResourceState();
+        if(!StringUtils.isNullOrEmpty(resourceModel.getArn())){
```
tyangjie, line 42: Use ```StringUtils.isNotEmpty```, and spacing

tyangjie (changes requested):

benkehoe:
Isn't it useful for resources to have a standardized "Arn" property, so as a user, I don't have to figure out what inevitably-inconsistent naming convention every particular resource provider decided to adopt for their Arn property? (I feel the name about "Name" as a property).

aws-cloudformation-resourcedefaultversion/aws-cloudformation-resourcedefaultversion.json
ugudip: Yup, The Arn in line 6 actually represents the TypeversionArn, I can change that.

aws-cloudformation-resourcedefaultversion/src/main/java/software/amazon/cloudformation/resourcedefaultversion/CreateHandler.java
ugudip: The reason for writing it this way is because, we need to first generate the TypeArn and set that PrimaryId generated to the resource model back in the progress call chain. This makes cfn decide if it's an already existing resource or a new one. If TypeArn is null, we generate one and return an IN_PROGRESS event. This then calls the remaining part of the logic(setTypeDefaultVersion) and completes after a read handler is called.

aws-cloudformation-resourcedefaultversion/src/main/java/software/amazon/cloudformation/resourcedefaultversion/CreateHandler.java
ugudip: Noted.

ugudip:
I made the following changes:
1. Made "Arn" as the Physical Id for ResourceDefaultVersion which represents the overall type(ex: arn:aws:cloudformation:us-west-2:336098619110:type/resource/Organization-Service-Resource11) and added it as a readOnly property
2. Changed the current "Arn" to "TypeVersionArn" which represents the Arn of the ResourceVersion. The user should now give "TypeVersionArn" in the template to refer ResourceVersion's Arn.
3. Kept the ResourceVersion's "Arn" as is because it represents the physical Id of the ResourceVersion and makes sense to keep it Arn.
3. Changed the Unit tests and code formatting.
4. Added a check to get the latest versionID in perdictArn() - bug fix

tyangjie (approved):

aws-cloudformation-resourcedefaultversion/src/main/java/software/amazon/cloudformation/resourcedefaultversion/CreateHandler.java
```diff
@@ -18,42 +12,44 @@
...
-            proxy.newInitiator(proxyClient, resourceModel, callbackContext);
-
-        logger.log(String.format("Creating [Arn: %s | Type: %s | Version: %s]",
-            resourceModel.getArn(), resourceModel.getTypeName(), resourceModel.getVersionId()));
-
-        return initiator
-            .translateToServiceRequest(Translator::translateToUpdateRequest) // same API call for Create + Update
-            .makeServiceCall((r, c) -> createResource(resourceModel, r, c))
-            .done(setTypeDefaultVersionResponse -> new ReadHandler().handleRequest(proxy, request, callbackContext, proxyClient, logger));
+        if (StringUtils.isNullOrEmpty(resourceModel.getArn())) {
```
jotompki, line 21: i feel like we should be using the callbackContext to set any sort of stateful things like the arn to indicate a transition. is the arn not being included in the initial input expected? what if someone includes this in their template? using the callbackContext to check if the arn has been set might avoid abusing the arn being set as some sort of state.

aws-cloudformation-resourcedefaultversion/src/main/java/software/amazon/cloudformation/resourcedefaultversion/CreateHandler.java
```diff
@@ -18,42 +12,44 @@
...
-        boolean exists = exists(proxyClient, model);
-        if (exists) {
-            this.logger.log("Default version for type is already set as requested");
-            throw new CfnAlreadyExistsException(ResourceModel.TYPE_NAME, model.getPrimaryIdentifier().toString());
+    private String createArn(ResourceHandlerRequest<ResourceModel> request) {
+        final ResourceModel resourceModel = request.getDesiredResourceState();
+        if (!StringUtils.isNullOrEmpty(resourceModel.getTypeVersionArn())) {
+            return resourceModel.getTypeVersionArn().substring(0, resourceModel.getTypeVersionArn().lastIndexOf("/"));
+        } else {
+            // generating Arn from the TypeName and versionID
```
jotompki, line 44: versionId isn't used to generate arn.

aws-cloudformation-resourcedefaultversion/src/main/java/software/amazon/cloudformation/resourcedefaultversion/CreateHandler.java
```diff
@@ -18,42 +12,44 @@
...
+            return ProgressEvent.progress(resourceModel, callbackContext);
+        }
+        return ProgressEvent.progress(resourceModel, callbackContext)
+                .then(progress -> {
+                    final ResourceModel model = progress.getResourceModel();
+                    logger.log(String.format("Creating [Arn: %s | Type: %s | Version: %s]",
+                            model.getTypeVersionArn(), model.getTypeName(), model.getVersionId()));
+                    return proxy.initiate("resourceDefaultVersion::Create", proxyClient, model, progress.getCallbackContext())
+                            .translateToServiceRequest(Translator::translateToUpdateRequest)
+                            .makeServiceCall((setTypeDefaultVersionRequest, client) -> proxyClient.injectCredentialsAndInvokeV2(setTypeDefaultVersionRequest, proxyClient.client()::setTypeDefaultVersion))
```
jotompki, line 33: do we map the TypeNotFoundException that's thrown here to the notfound exception in the handler? if  a resource type doesnt exist and we try to set a default version, what happens?

aws-cloudformation-resourcedefaultversion/src/main/java/software/amazon/cloudformation/resourcedefaultversion/CreateHandler.java
ugudip: Sure, I will change the comment here, I actually meant when TypeName and VersionId are given as an input..

aws-cloudformation-resourcedefaultversion/src/main/java/software/amazon/cloudformation/resourcedefaultversion/CreateHandler.java
ugudip: Made changes.

aws-cloudformation-resourcedefaultversion/src/main/java/software/amazon/cloudformation/resourcedefaultversion/CreateHandler.java
ugudip: In that case, the resource creation fails anyway. But I added the check in the handle error clause to let Cfn know it's a NotFoundException.

jotompki (approved):

PR #29, opened 2020-11-24
Ugudip resource version call chain branch

ugudip:
Changes:

1. Removed the callchain initiator and rebind model changes and replaced it with simple progress chain logic to prevent call chain being overused in the predict Arn method.
2. Added Unit tests to the made changes.

### DIFF
--- a/aws-cloudformation-resourceversion/src/main/java/software/amazon/cloudformation/resourceversion/CreateHandler.java
+++ b/aws-cloudformation-resourceversion/src/main/java/software/amazon/cloudformation/resourceversion/CreateHandler.java
@@ -27,14 +27,16 @@ public class CreateHandler extends BaseHandlerStd {
 
         ResourceModel resourceModel = request.getDesiredResourceState();
         return ProgressEvent.progress(resourceModel, callbackContext)
+                .then(progress -> predictArn(progress, proxyClient, request))
                 .then(progress ->
                         proxy.initiate("create", proxyClient, resourceModel, callbackContext)
                                 .translateToServiceRequest(Translator::translateToCreateRequest)
                                 .makeServiceCall((awsRequest, sdkProxyClient) -> sdkProxyClient.injectCredentialsAndInvokeV2(awsRequest, sdkProxyClient.client()::registerType))
                                 .done((registerTypeRequest, registerTypeResponse, sdkProxyClient, model, cc) -> {
-                                    cc.setRegistrationToken(registerTypeResponse.registrationToken());
-                                    return predictArn(progress, proxyClient, request);
-                                })
+                                            cc.setRegistrationToken(registerTypeResponse.registrationToken());
+                                            return ProgressEvent.progress(model, cc);
+                                        }
+                                )
                 )
                 .then(progress -> stabilizeOnCreate(progress, proxy, proxyClient))
                 .then(progress -> new ReadHandler().handleRequest(proxy, request, progress.getCallbackContext(), proxyClient, logger));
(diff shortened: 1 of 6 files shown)


aws-cloudformation-resourceversion/src/main/java/software/amazon/cloudformation/resourceversion/CreateHandler.java
```diff
@@ -47,88 +45,77 @@
...
-            if (!response.isSuccess()) {
-                return ProgressEvent.failed(
-                        model, context, response.getErrorCode(), response.getMessage());
+            //make a ListTypeVersions API call to get all versionIDs for status LIVE for the first time
+            final ListTypeVersionsResponse listTypeVersionsResponse;
+
+            try {
+                listTypeVersionsResponse = proxyClient.injectCredentialsAndInvokeV2(
+                        Translator.translateToListTypeVersionsRequest(model, marker, context.getDeprecatedStatus()), proxyClient.client()::listTypeVersions);
+            } catch (CfnRegistryException e) {
```
ammokhov, line 63: what does this exception mean?

aws-cloudformation-resourceversion/src/main/java/software/amazon/cloudformation/resourceversion/CreateHandler.java
```diff
@@ -47,88 +45,77 @@
...
-                        } else {
-                            return latestVersion < first ? summary : currentLatest;
-                        }
-                    }
-            ).orElse(latest);
-            if (callResponse.nextToken() != null) {
-                listApi = listApi.rebindModel(callResponse);
+            latest = getLatestVersion(latest, currentLatest, listTypeVersionsResponse);
+            marker = listTypeVersionsResponse.nextToken();
+            if (StringUtils.isNullOrEmpty(marker) && context.getDeprecatedStatus() == DeprecatedStatus.LIVE) {
```
ammokhov, line 77: if token is null it means the pages have been exhausted

aws-cloudformation-resourceversion/src/main/java/software/amazon/cloudformation/resourceversion/CreateHandler.java
```diff
@@ -47,88 +45,77 @@
...
             }
-
-            String arn = latest.arn();
-            String currentArn = currentLatest.arn();
-            Integer currentLatestVersion = Integer.valueOf(currentArn.substring(currentArn.lastIndexOf("/") + 1));
-            Integer latestVersion = Integer.valueOf(arn.substring(arn.lastIndexOf("/") + 1));
-            Integer current = currentLatestVersion > latestVersion ? currentLatestVersion : latestVersion; // to get the latest(largest) version ID
-            arn = arn.substring(0, arn.lastIndexOf("/") + 1)
-                    .concat(String.format("%08d", current + 1));
+            String arn = getCurrentArn(latest, currentLatest); //returns the largest version ID +1
```
ammokhov, line 82: not sure Im following why u need this

aws-cloudformation-resourceversion/src/main/java/software/amazon/cloudformation/resourceversion/CreateHandler.java
ugudip: The ListTypeVersions API throws CfnRegistryException when the type is not yet registered. In this case the predicted version id will be 0000001.

aws-cloudformation-resourceversion/src/main/java/software/amazon/cloudformation/resourceversion/CreateHandler.java
ugudip: yes, but we need to check for each status - LIVE and DEPRECATED. The default is LIVE and if the token is exhausted for LIVE, we need to switch it to DEPRECATED and check for the latest verison ID.

aws-cloudformation-resourceversion/src/main/java/software/amazon/cloudformation/resourceversion/CreateHandler.java
ugudip: This method decides the final predicted Arn of the version. The latest and the currentLatest both are the maximum versions for Live and deprecated status respectively. We need to extract the Arn from these TypeVersionSummaries by picking the largest version and incrementing 1.

PR #34, opened 2020-12-09
Setting ModuleVersion's ModulePackage field to "required"

kristian-d:
I had originally left `ModuleVersion`'s `ModulePackage` field out of the "required" fields in the schema because it was causing issues with contract tests (i.e. contract tests would fail when attempting a delete operation because the module package field was required, but cannot be read in `DescribeType` requests and supplied to the delete handler [#318](https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/issues/318)). I had temporarily implemented manual validation in the `ModuleVersion` create handler to check for `ModulePackage` to compensate.

After building, running the unit tests, and testing these changes manually in my account through a number of create-, update-, and delete-stack operations, I am confident setting `ModulePackage` as "required" is safe, despite being flagged by the contract tests for the above reason. `ModulePackage` is not required as input to the delete handler -- only the ARN is.

With this change, the `AWS::CloudFormation::ModuleVersion` documentation will now agree with the schema as well.

I also added `ModuleName` to `createOnlyProperties`, as it should have been.

### DIFF
--- a/aws-cloudformation-moduleversion/src/main/java/software/amazon/cloudformation/moduleversion/CreateHandler.java
+++ b/aws-cloudformation-moduleversion/src/main/java/software/amazon/cloudformation/moduleversion/CreateHandler.java
@@ -125,14 +125,4 @@ public class CreateHandler extends BaseHandlerStd {
                 throw new CfnGeneralServiceException(String.format("received unexpected module registration status: %s", dtrResponse.progressStatus()));
         }
     }
-
-    @Override
-    protected void validateModel(final ResourceModel model) {
-        if (model == null) {
-            throw new CfnInvalidRequestException("ResourceModel is required");
-        }
-        if (model.getModulePackage() == null) {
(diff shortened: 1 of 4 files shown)


aws-cloudformation-moduleversion/src/main/java/software/amazon/cloudformation/moduleversion/CreateHandler.java
```diff
@@ -125,14 +125,4 @@ private Boolean stabilize(
                 throw new CfnGeneralServiceException(String.format("received unexpected module registration status: %s", dtrResponse.progressStatus()));
         }
     }
-
-    @Override
-    protected void validateModel(final ResourceModel model) {
```
priyanka-amzn, line 130: If you still have the check for ModuleName in code, this UT will improve the coverage of the handler.

priyanka-amzn (approved):

miparnisari (approved):

ugudip (approved):

PR #37, opened 2020-12-18
StackSet ReadHandler populates OUIDs

xiwhuang:
*Issue #, if available:*

*Description of changes:*

* ReadHandler gets OuIds when using Service Managed StackSets as Describe API now returns Orgs Info
* ~Using `runWithRetry` method to call Describe to avoid throttling.~

### Testing
1. ~Unit test 100% covers `RetryUtils` on both lines and branches.~
2. ~Manual tests calling ReadHandler with (200 / 60 = ) 3.33 TPS under 20 threads. 15 % failed due to reaching max 3 attempts. However, positive with the result as normal use case would be usually only one concurrency, P99 cases should succeed after one retry (500 ms sleep time) as default throttling TPS limit is 2.~
3. Resource Import Service Managed StackSets Succeeded.

### R2
* Avoid to calling DescribeStackInstance in Read, as ReadHandler will only return partial properties for now.

### R3
* `ResourceModelBuilder` should handle properly when `PermissionModel` is `null`


aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/util/RetryUtils.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.util.function.Supplier;
+
+@RequiredArgsConstructor
+public class RetryUtils {
+    private final static int MAX_RETRIES = 3;
+    private final Logger logger;
+    private int retryCount = 0;
+
+    RetryUtils(Logger logger, int retryCount) {
```
guanghug: Seems this is only used for testing, you'd better avoid adding this constructor because passing retryCount is not a normal case, and retryCount is more like an internal function variable.

guanghug (approved):

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/util/RetryUtils.java
xiwhuang: Resolving as the `RetryUtils` was removed

wbingli (approved):

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/util/ResourceModelBuilder.java
```diff
@@ -87,28 +83,24 @@ public ResourceModel buildModel() {
...
-        if (!listStackInstancesResponse.hasSummaries()) return;
+        final String nextToken = listStackInstancesResponse.nextToken();
+        if (!listStackInstancesResponse.hasSummaries()) return null;
         listStackInstancesResponse.summaries().forEach(member -> {
-            final List<Parameter> parameters = getStackInstance(member);
-            stackInstanceSet.add(translateToStackInstance(isSelfManaged, member, parameters));
+            // Parameters are set null as we can't retrieve parameter override from List API.
+            // Retrieving from Describe API requires to brutal force every single stack instance
+            // which will likely cause timeout issue
+            stackInstanceSet.add(translateToStackInstance(isSelfManaged, member, null));
```
sungkkim, line 102: Just for future reference, this will likely break drift detection. We will have to return the full model before enabling drift detection on this resource type.

aws-cloudformation-stackset/src/test/java/software/amazon/cloudformation/stackset/ReadHandlerTest.java
```diff
@@ -82,6 +72,29 @@ public void handleRequest_SelfManagedSS_Success() {
 
         verify(proxyClient.client()).describeStackSet(any(DescribeStackSetRequest.class));
         verify(proxyClient.client()).listStackInstances(any(ListStackInstancesRequest.class));
-        verify(proxyClient.client(), times(4)).describeStackInstance(any(DescribeStackInstanceRequest.class));
+    }
+
+    @Test
+    public void handleRequest_PermissionModelIsNull() {
```
sungkkim, line 78: 👍

sungkkim (approved):

PR #41, opened 2021-01-15
Remove predictArn from the ResourceVersion CreateHandler, add TypeArn…

ugudip:
… in the schema files and other unit test changes

Below are the changes in this PR:

1. Removed the predictArn method in ResourceVersion CreateHandler. This prediction was causing a lot of throttling issues and prediction might take > 1 min for larger versionIDs. We can now have TypeVersionArn in the DescribeTypeRegistration  response immediately, so we need not predict the TypeVersionArn anymore.
2. Added TypeArn in the ResourceVersion schema file and made TypeVersionArn as the primary identifier instead of just "Arn". 
3. Made TypeArn as the primary Identifier for ResourceDefaultVersion instead of just "Arn"
4. Added respective unit tests
5. Did some code reformat
6. Bumped up the versions in pom.xml

Contract test result:

1. ResourceVersion
handler_create.py::contract_create_delete `PASSED`  
handler_create.py::contract_invalid_create `PASSED`
handler_create.py::contract_create_duplicate `SKIPPED` (No writable identifiers. Skipping test.) 
handler_create.py::contract_create_read_success `PASSED`
handler_create.py::contract_create_list_success `PASSED`
handler_delete.py::contract_delete_read `PASSED`
handler_delete.py::contract_delete_list `PASSED`
handler_delete.py::contract_delete_delete `PASSED`
handler_delete.py::contract_delete_create `SKIPPED` (No writable identifiers. Skipping test.) 
handler_misc.py::contract_check_asserts_work `PASSED`
handler_read.py::contract_read_without_create `PASSED` 
 
 The above result was seen after removing the schemaHandlerPackage from writeOnlyProperties. I raised a bug in contract 
 tests here https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/issues/318

2. ResourceDefaultVersion
handler_create.py::contract_create_delete `PASSED`
handler_create.py::contract_invalid_create `PASSED` 
handler_create.py::contract_create_duplicate `SKIPPED` (No writable identifiers. Skipping test.)
handler_create.py::contract_create_read_success `PASSED`
handler_delete.py::contract_delete_read `FAILED`
handler_delete.py::contract_delete_update `FAILED` 
handler_delete.py::contract_delete_delete `FAILED`
handler_delete.py::contract_delete_create `SKIPPED` (No writable identifiers. Skipping test.)
handler_misc.py::contract_check_asserts_work `PASSED`
handler_read.py::contract_read_without_create `PASSED`
handler_update.py::contract_update_read_success `PASSED`
handler_update_invalid.py::contract_update_create_only_property SKIPPED (No createOnly Properties. Skipping test.)

 The reason for the above 3 FAILED tests is because the Delete handler for this resource is a no-op, hence we actually can't 
 "delete" a default version, we can only change it. The above 3 contract tests ensure the resource is deleted after a delete 
 handler, hence they fail for this resource.


aws-cloudformation-resourceversion/aws-cloudformation-resourceversion.json
```diff
@@ -118,23 +123,24 @@
     "/properties/DocumentationUrl",
     "/properties/LastUpdated",
```
wbingli, line 124: I don't think the LastUpdated and TimeCreated has any value for customer to get. Would be good to remove from the model to keep it simple.

aws-cloudformation-resourceversion/src/main/java/software/amazon/cloudformation/resourceversion/CreateHandler.java
```diff
@@ -1,167 +1,61 @@
...
+                                .done((registerTypeRequest, registerTypeResponse, sdkProxyClient, model, cc) -> {
+                                            cc.setRegistrationToken(registerTypeResponse.registrationToken());
+                                            DescribeTypeRegistrationResponse describeTypeRegistrationResponse =
+                                                    proxy.injectCredentialsAndInvokeV2(Translator
+                                                                    .translateToDescribeTypeRegistration(cc.getRegistrationToken()),
+                                                            proxyClient.client()::describeTypeRegistration);
+                                            if (describeTypeRegistrationResponse == null) {
+                                                throw new CfnInternalFailureException();
+                                            }
+                                            model.setTypeVersionArn(describeTypeRegistrationResponse.typeVersionArn());
```
wbingli, line 39: nice.

aws-cloudformation-resourceversion/aws-cloudformation-resourceversion.json
ugudip: That's true but those properties are read only and are not set by the user. These properties are returned in the Read handler in order to mirror the underlying DescribeType API. So I think it's ok to have them in the schema.

wbingli: I don't think we need to mirror all the properties from the low level service API like DescribeType.   Those information are most not useful to cloudformation resource.  

Therefore, in most of the case, we should remove those properties to keep the resource simple.

It's easier to add those properties if there is a customer use case. But it will be hard to change if we release at the beginning.

The idea for readOnly properties is to only for the properties which are absolutely critical, such as Arn, server generate ID.

ugudip: I see. Removed these ReadOnly properties from the schema - SourceUrl, DocumentationUrl, Schema, Description, LastUpdated, TimeCreated.

wbingli (approved):

jotompki (approved):

PR #44, opened 2021-02-02
Remove ARN prediction from AWS::CloudFormation::ModuleVersion

kristian-d:
This PR removes ARN prediction from the *AWS::CloudFormation::ModuleVersion* resource provider; the ARN can now be retrieved from the initial `DescribeTypeRegistrationResponse` and prediction is no longer needed for returning a primary identifier.

Added more information to logs and cleaned logging up.

Unit tests are passing and unit tests were added for new cases. Contract test results remain the same as previously.

- handler_create.py::contract_create_delete PASSED
- handler_create.py::contract_invalid_create PASSED
- handler_create.py::contract_create_duplicate SKIPPED (No writable identifiers. Skipping test.)
- handler_create.py::contract_create_read_success PASSED
- handler_create.py::contract_create_list_success PASSED
- handler_delete.py::contract_delete_read PASSED
- handler_delete.py::contract_delete_list PASSED
- handler_delete.py::contract_delete_delete PASSED
- handler_delete.py::contract_delete_create SKIPPED (No writable identifiers. Skipping test.)
- handler_misc.py::contract_check_asserts_work PASSED
- handler_read.py::contract_read_without_create PASSED


aws-cloudformation-moduleversion/.rpdk-config
```diff
@@ -13,5 +14,6 @@
         ],
         "codegen_template_path": "guided_aws",
         "protocolVersion": "2.0.0"
-    }
+    },
+    "executableEntrypoint": "software.amazon.cloudformation.moduleversion.HandlerWrapperExecutable"
```
priyanka-amzn, line 18: What does this change do?

aws-cloudformation-moduleversion/src/main/java/software/amazon/cloudformation/moduleversion/CallbackContext.java
```diff
@@ -16,7 +16,7 @@
 
     @Getter
```
priyanka-amzn, line 17: This `@Getter` and `@Setter` for the parameter can be removed too.

aws-cloudformation-moduleversion/src/main/java/software/amazon/cloudformation/moduleversion/CreateHandler.java
```diff
@@ -69,12 +59,31 @@ public CreateHandler() {
...
+    private RegisterTypeResponse registerModule(
+            final RegisterTypeRequest request,
+            final ProxyClient<CloudFormationClient> proxyClient,
+            final ResourceModel model) {
         RegisterTypeResponse response;
         try {
             response = proxyClient.injectCredentialsAndInvokeV2(request, proxyClient.client()::registerType);
         } catch (final CfnRegistryException exception) {
-            logger.log(String.format("Failed to register module:\n%s", Arrays.toString(exception.getStackTrace())));
+            logger.log(String.format("Failed to register new version of module %s:\n%s",
```
priyanka-amzn, line 70: Would this always be a new version? What about the first version?

aws-cloudformation-moduleversion/src/main/java/software/amazon/cloudformation/moduleversion/CreateHandler.java
```diff
@@ -89,27 +98,24 @@ private Boolean stabilize(
...
-        DescribeTypeRegistrationResponse dtrResponse;
-        try {
-            dtrResponse = proxyClient.injectCredentialsAndInvokeV2(dtrRequest, proxyClient.client()::describeTypeRegistration);
-        } catch (final CfnRegistryException exception) {
-            logger.log(String.format("Failed to retrieve module registration status for module %s:\n%s",
-                    model.getModuleName(), Arrays.toString(exception.getStackTrace())));
-            throw new CfnGeneralServiceException(exception);
+        final String typeVersionArn = dtrResponse.typeVersionArn();
+        if (typeVersionArn != null) {
+            if (model.getArn() != null && !model.getArn().equals(typeVersionArn)) {
```
priyanka-amzn, line 106: Nice :)

aws-cloudformation-moduleversion/.rpdk-config
```diff
@@ -1,4 +1,5 @@
 {
+    "artifact_type": "RESOURCE",
```
priyanka-amzn, line 2: Was `artifact_type` not required earlier?

aws-cloudformation-moduleversion/src/main/java/software/amazon/cloudformation/moduleversion/CallbackContext.java
kristian-d: I've left this field in `CallbackContext` because it is used by the `ListHandler`. My previous comment, "can be removed once ArnPredictor.java is removed," was incorrect.

aws-cloudformation-moduleversion/src/main/java/software/amazon/cloudformation/moduleversion/CreateHandler.java
kristian-d: When `ModuleVersion`'s `CreateHandler` is invoked and this `registerModule` method is called, it is always attempting to register a new version under the specified module name. I don't quite understand your question about the first version, though.

aws-cloudformation-moduleversion/.rpdk-config
kristian-d: The changes in this file weren't my doing, but a new automated addition when running, I believe, `cfn validate`. I kept the changes to stay consistent with `ResourceVersion`. See [here](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-cloudformation/pull/41/files#diff-89ed9acf8e311f86afbc53ffa55f49aafba6a1c976da5e6c768048a36c1bac08).

aws-cloudformation-moduleversion/.rpdk-config
kristian-d: Please see above comment.

In answer to your question, I am not 100% sure, but given that this is `.rpdk-config`, I expect that this is used by the cloudformation-cli during local execution.

aws-cloudformation-moduleversion/src/main/java/software/amazon/cloudformation/moduleversion/CallbackContext.java
priyanka-amzn: I misread the file diff. Got it.

aws-cloudformation-moduleversion/src/main/java/software/amazon/cloudformation/moduleversion/CreateHandler.java
priyanka-amzn: When the module is registered for the first time that would be the first version of the module. `Failed to register new version of module` would be a confusing error message in that case. 
Every time the CreateHandler is called, a new version gets created and that's fine. `New version` indicates in a way that there was an older version which is confusing when the first version is being registered.

aws-cloudformation-moduleversion/src/main/java/software/amazon/cloudformation/moduleversion/CallbackContext.java
priyanka-amzn: I misread the diff, sorry.

aws-cloudformation-moduleversion/src/main/java/software/amazon/cloudformation/moduleversion/CreateHandler.java
kristian-d: I've updated the logging now. It should be less confusing.

priyanka-amzn (approved):

aws-cloudformation-moduleversion/.rpdk-config
ugudip: Yes, the above two keys were added from the new rpdk version.

aws-cloudformation-moduleversion/src/main/java/software/amazon/cloudformation/moduleversion/CreateHandler.java
ugudip: Just wondering, don't we need to handle this here instead of a logger? If you have the arn set to the model already and the typeVersionArn now returned is a different one, which version is the module currently linked to?

kristian-d: In my experience, CloudFormation assigns the first primary identifier it receives to a resource and, if the primary identifier changes during resource stabilization, CloudFormation does not reassign the resource. In this case, you are right and there would be unexpected behaviour; the stack would track a different ARN than the registry. I will update this to throw a `CfnGeneralServiceException` exception.

ugudip (approved):

priyanka-amzn (approved):

jotompki (approved):

Issue #49, opened 2021-02-22
Adding new module/module default version in same template fails if module already existed


tylersouthwick:
If module was already defined, but all versions were deregistered, adding a new module/module default version in the same template fails.

It seems to fail due to this check: https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-cloudformation/blob/master/aws-cloudformation-moduledefaultversion/src/main/java/software/amazon/cloudformation/moduledefaultversion/CreateHandler.java#L42

I'm assuming that check is there to prevent an implicit "import" of a different default version?

tylersouthwick:
The issue we've run into is that in this case, you are unable to use the same template to deploy a new version of the module. You have to manage the state external to cloudformation and know whether the stack deployment is the first deployment or the second. This makes using stacksets for multi account deployment really difficult.

We have a current workaround, but it is really clunky. The example on https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cloudformation-moduledefaultversion.html shows `2` different modules defined, but setting the default version only on `ModuleVersion2`. `ModuleVersion1` is essentially only used to define the existence of the module. Ideally, we could have a single template that defines only a `ModuleVersion` and a `ModuleVersion` that can then be updated when the module changes and can also be used for the initial deployment.

priyanka-amzn:
Hi Tyler, thank you for your patience. We're sorry you have to workaround this issue. We're having discussions with CloudFormation product managers, engineering managers and owners of `ResourceDefaultVersion` to discuss the path forward for `ModuleDefaultVersion`. 

We understand that removing `AlreadyExists` in this scenario would be better customer experience. We'll get back to you on this.

priyanka-amzn:
Hi Tyler, our deployment is complete. Setting a `ModuleDefaultVersion` in a template to a version that is already the default version should not report an error now. Thanks for your patience.

PR #50, opened 2021-02-24
do not check if module default version exists

tylersouthwick:
*Issue #, if available:*
#49 

*Description of changes:*
I had looked at checking if the default version is the _same_ as the one that we're trying to set. But it seems that that the API doesn't throw an error in that case. The only issue I can see is that this could allow a different stack to manage the same default version resource.

[ResourceDefaultVersion ](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-cloudformation/blob/master/aws-cloudformation-resourcedefaultversion/src/main/java/software/amazon/cloudformation/resourcedefaultversion/CreateHandler.java)does not have this check.

### DIFF
--- a/aws-cloudformation-moduledefaultversion/src/test/java/software/amazon/cloudformation/moduledefaultversion/CreateHandlerTest.java
+++ b/aws-cloudformation-moduledefaultversion/src/test/java/software/amazon/cloudformation/moduledefaultversion/CreateHandlerTest.java
@@ -3,13 +3,9 @@ package software.amazon.cloudformation.moduledefaultversion;
 import org.mockito.ArgumentCaptor;
 import software.amazon.awssdk.services.cloudformation.CloudFormationClient;
 import software.amazon.awssdk.services.cloudformation.model.CfnRegistryException;
-import software.amazon.awssdk.services.cloudformation.model.DeprecatedStatus;
-import software.amazon.awssdk.services.cloudformation.model.DescribeTypeRequest;
-import software.amazon.awssdk.services.cloudformation.model.DescribeTypeResponse;
 import software.amazon.awssdk.services.cloudformation.model.SetTypeDefaultVersionRequest;
 import software.amazon.awssdk.services.cloudformation.model.SetTypeDefaultVersionResponse;
 import software.amazon.awssdk.services.cloudformation.model.TypeNotFoundException;
-import software.amazon.cloudformation.exceptions.CfnAlreadyExistsException;
 import software.amazon.cloudformation.exceptions.CfnGeneralServiceException;
 import software.amazon.cloudformation.exceptions.CfnInvalidRequestException;
 import software.amazon.cloudformation.exceptions.CfnNotFoundException;
@@ -18,7 +14,6 @@ import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.junit.jupiter.MockitoExtension;
 import software.amazon.cloudformation.proxy.AmazonWebServicesClientProxy;
-import software.amazon.cloudformation.proxy.HandlerErrorCode;
 import software.amazon.cloudformation.proxy.Logger;
 import software.amazon.cloudformation.proxy.OperationStatus;
 import software.amazon.cloudformation.proxy.ProgressEvent;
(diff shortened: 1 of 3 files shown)


priyanka-amzn:
Hello, 
This PR proposes to remove the `AlreadyExists` check that we put in place for consistency keeping expected behavior for all Registry resource-types in mind.

More details on the check:
- Registry contract tests require that, if a resource already exists, then AlreadyExists is thrown
- It was decided that, if a customer were to set a default version to version `x`, then attempt to set version `x` as the default version again, returning `AlreadyExists` exception is consistent with other resource types.

priyanka-amzn:
When a user uses a `ModuleVersion` and `ModuleDefaultVersion` in the same template after de-registering all the versions of this `ModuleVersion` earlier, the user can include only `ModuleVersion` in the template. `ModuleDefaultVersion` would not be required. Including it will return an error because it tries to set a version that is already the default version as the default version.

This is because if the CloudFormation Registry service gets a request to create a new `ModuleVersion` and there is no existing default version for it,  it sets that version as the default version.

`ModuleDefaultVersion` is to be used when there are multiple versions and the customer wants to change the default version.

tylersouthwick:
@priyanka-amzn please see the comment I added in #49

priyanka-amzn:
Thanks for your patience Tyler. We've discussed this with managers and decided that removing the `AlreadyExists` exception in `ModuleDefaultVersion` would be better for customers and this would also make it consistent with `ResourceDefaultVersion`'s behavior.

@kristian-d and I will review this PR. Thanks again for your contribution.

aws-cloudformation-moduledefaultversion/src/test/java/software/amazon/cloudformation/moduledefaultversion/CreateHandlerTest.java
```diff
@@ -28,10 +23,7 @@
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.mockito.ArgumentMatchers.any;
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
priyanka-amzn, line 26: It would be great if you could replace this wildcard import(`*`) with specific imports. 
https://stackoverflow.com/questions/147454/why-is-using-a-wild-card-with-a-java-import-statement-bad

priyanka-amzn (approved):

priyanka-amzn:
@kristian-d please review.

aws-cloudformation-moduledefaultversion/src/test/java/software/amazon/cloudformation/moduledefaultversion/CreateHandlerTest.java
```diff
@@ -152,157 +146,45 @@ public void handleRequest_Translation_NameAndVersionInput() {
...
-        verify(readHandler, times(1))
-                .handleRequest(any(AmazonWebServicesClientProxy.class), any(), any(CallbackContext.class), any(), any(Logger.class));
-        verify(client, times(0)).setTypeDefaultVersion(any(SetTypeDefaultVersionRequest.class));
+        verify(client, times(1)).setTypeDefaultVersion(any(SetTypeDefaultVersionRequest.class));
     }
 
     @Test
-    public void handleRequest_SetDefaultVersion_NotFound() {
+    public void handleRequest_SetDefaultVersion_NotFound_ModuleNameVersionIdInput() {
         final ResourceModel modelIn = ResourceModel.builder()
```
kristian-d, line 171: Could you please update this to use 
`.moduleName(moduleName)` and `.versionId(versionId);`?

kristian-d:
Looks good! I just left one minor comment on a unit test.

aws-cloudformation-moduledefaultversion/src/test/java/software/amazon/cloudformation/moduledefaultversion/CreateHandlerTest.java
tylersouthwick: done

priyanka-amzn (approved):

tylersouthwick:
@kristian-d I addressed the comment.

kristian-d (approved):

PR #61, opened 2021-04-21
Add callAs parameter to StackSet Resource

aradven:
*Issue #, if available:*

*Description of changes:*
Adds the callAs parameter to the StackSet resource, enabling use of delegated administration

### DIFF
--- a/aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/ReadHandler.java
+++ b/aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/ReadHandler.java
@@ -1,6 +1,10 @@
 package software.amazon.cloudformation.stackset;
 
+import software.amazon.awssdk.awscore.exception.AwsServiceException;
 import software.amazon.awssdk.services.cloudformation.CloudFormationClient;
+import software.amazon.awssdk.services.cloudformation.model.CallAs;
+import software.amazon.awssdk.services.cloudformation.model.StackSet;
+import software.amazon.awssdk.services.cloudformation.model.StackSetNotFoundException;
 import software.amazon.cloudformation.proxy.AmazonWebServicesClientProxy;
 import software.amazon.cloudformation.proxy.Logger;
 import software.amazon.cloudformation.proxy.OperationStatus;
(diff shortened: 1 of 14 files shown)


aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/ReadHandler.java
```diff
@@ -23,8 +23,9 @@
         return ProgressEvent.<ResourceModel, CallbackContext>builder()
                 .resourceModel(ResourceModelBuilder.builder()
                         .proxyClient(proxyClient)
-                        .stackSet(describeStackSet(proxyClient, model.getStackSetId()))
-                        .build().buildModel())
+                        .stackSet(describeStackSet(proxyClient, model.getStackSetId(), model.getCallAs()))
+                        .build()
+                        .buildModel(model.getCallAs()))
```
xiwhuang, line 28: I am afraid, we might not be able to `getCallAs` in read handler as it is not one of the identifiers.

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/util/ResourceModelBuilder.java
```diff
@@ -32,12 +32,15 @@
...
+    public ResourceModel buildModel() {
+        return buildModel(null);
+    }
     /**
      * Returns the model we construct from StackSet service client using PrimaryIdentifier StackSetId
      *
      * @return {@link ResourceModel}
      */
-    public ResourceModel buildModel() {
+    public ResourceModel buildModel(final String callAs) {
```
xiwhuang, line 43: Nit: We can actually add `CallAs` as a member variable to be consistent.

aradven: I understand why, but I think I'd rather not here. CallAs is not really part of the actual model, it's part of the request, and I think we should reflect that where possible

xiwhuang (approved):
LGTM, just to make sure it works for resource import as well

aws-cloudformation-stackset/src/test/java/software/amazon/cloudformation/stackset/CreateHandlerTest.java
```diff
@@ -99,6 +102,45 @@ public void handleRequest_ServiceManagedSS_SimpleSuccess() {
         verify(proxyClient.client()).describeStackSetOperation(any(DescribeStackSetOperationRequest.class));
     }
 
+    @Test
+    public void handleRequest_ServiceManagedSS_WithCallAs_SimpleSuccess() {
```
xalle-y, line 106: Any reason not to include tests for SelfManaged with CallAs?

aradven: Mostly cause there's no functional difference within the handler's themselves. It's just another parameter that's passed through to the SDK, we don't do logic with it here.

xalle-y: We should not be assuming this type of behaviour when deciding what should or should not be tested. The main addition here is "CallAs for Self and Service-managed stack sets" so expected behaviour for those cases should be tested. Even if it's a no-op just having the test there is a better way to solidify that expectation.

aradven: Fair enough, I'll add some

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/util/ResourceModelBuilder.java
aradven: Realized that import would actually probably rely on this existing, so adding it to the model

xalle-y (approved):

PR #65, opened 2021-04-26
SkyTrain-NS-259 Bugfix: Ensure same client request token is used when register-type is resent

MalikAtalla-AWS:
The CreateHandler is being called multiple times by the CloudFormation framework (documented for example [here](https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-walkthrough.html)), so we have to make sure it's idempotent. To do that we ensure the same clientRequestToken is sent along with the register type call. The current bug is triggered because the random clientRequestTokens cause the hash of the RegisterTypeRequest to be different between CreateHandler invokation which causes CFN to resend the request (see [hashing of the request here](https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/blob/master/src/main/java/software/amazon/cloudformation/proxy/AmazonWebServicesClientProxy.java#L265))

### DIFF
--- a/aws-cloudformation-moduleversion/src/main/java/software/amazon/cloudformation/moduleversion/CreateHandler.java
+++ b/aws-cloudformation-moduleversion/src/main/java/software/amazon/cloudformation/moduleversion/CreateHandler.java
@@ -1,5 +1,6 @@
 package software.amazon.cloudformation.moduleversion;
 
+import lombok.NonNull;
 import software.amazon.awssdk.services.cloudformation.CloudFormationClient;
 import software.amazon.awssdk.services.cloudformation.model.CfnRegistryException;
 import software.amazon.awssdk.services.cloudformation.model.DescribeTypeRegistrationRequest;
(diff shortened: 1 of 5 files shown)


aws-cloudformation-moduleversion/src/main/java/software/amazon/cloudformation/moduleversion/CreateHandler.java
```diff
@@ -58,6 +62,17 @@ public CreateHandler() {
                 .then(progress -> readHandler.handleRequest(proxy, request, callbackContext, proxyClient, logger));
     }
 
+    private String getOrGenerateClientRequestToken() {
+      if (this.callbackContext.getClientRequestToken() == null) {
```
kristian-d, line 66: Nit: I can't quite tell, but I think the indentation of this method is different than the remainder of the file.

MalikAtalla-AWS: Yeah, you're right. Fixed.

kristian-d (approved):

miparnisari (approved):

PR #74, opened 2021-09-13
StackSets - ListStackSetsRequest with ACTIVE status

xiwhuang:
*Issue #, if available:*

Preivously, we list all stacksets including `Deleted` Stacksets, that causes an unnecessary latency issue to users who have a lot of StackSets deleted before.

*Description of changes:*

Optimizing the List handler to call `ListStackSets` with `ACTIVE` status flag

### DIFF
--- a/.github/workflows/java-ci-with-maven.yml
+++ b/.github/workflows/java-ci-with-maven.yml
@@ -30,7 +30,7 @@ jobs:
     - name: Install jinja2 with 2.11.3 to resolve incompatible issue
       run: pip install jinja2==2.11.3
     - name: Set up cloudformation-cli-java-plugin
-      run: pip install cloudformation-cli-java-plugin
+      run: pip install cloudformation-cli-java-plugin==2.0.3
     - name: install and run pre-commit
       uses: pre-commit/action@v2.0.0
       with:
--- a/aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/translator/RequestTranslator.java
+++ b/aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/translator/RequestTranslator.java
@@ -18,6 +18,7 @@ import software.amazon.cloudformation.stackset.StackInstances;
 
 import java.util.Map;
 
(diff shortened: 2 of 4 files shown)


aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/translator/RequestTranslator.java
```diff
@@ -124,6 +125,7 @@ public static ListStackSetsRequest listStackSetsRequest(
         return ListStackSetsRequest.builder()
                 .maxResults(LIST_MAX_ITEMS)
                 .nextToken(nextToken)
+                .status(ACTIVE)
```
xalle-y, line 128: Are there any places where we may want to list deleted stacksets?

xalle-y (approved):

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/translator/RequestTranslator.java
xiwhuang: Not that I know, at least, currently the list stacksets are only used in list handler. which will just return a list of live StackSets

ugudip:
Can we add more context on the description on what this PR is doing? Is this to list only active stacksets every time?

.github/workflows/java-ci-with-maven.yml
```diff
@@ -30,7 +30,7 @@ jobs:
     - name: Install jinja2 with 2.11.3 to resolve incompatible issue
       run: pip install jinja2==2.11.3
     - name: Set up cloudformation-cli-java-plugin
-      run: pip install cloudformation-cli-java-plugin
+      run: pip install cloudformation-cli-java-plugin==2.0.3
```
ugudip, line 33: Spoke to @xiwhuang offline, the current java plugin is breaking the build checks, so downgrading it to 2.0.3 for now. However, we should revert this change once the latest version is working as expected because it's recommended to use the latest version.

xiwhuang:
Sure. Not sure why the description was not added. Basically it is optimizing the impl in List handler.

ugudip (approved):

PR #104, opened 2022-12-21
add support for nested stacks in self managed stack sets

macdmerr:
*Issue #, if available:*

*Description of changes:*
Stack Sets supports nested stacks for self managed stack sets.  in order to support feature parity, the resource handler should be updated to match that behavior. 

### DIFF
--- a/aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/util/Validator.java
+++ b/aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/util/Validator.java
@@ -24,18 +24,27 @@ public class Validator {
     private final static String DIFF = "DIFFERENCE";
     private final static String NONE = "NONE";
     private final static HashSet<String> validAccountFilters = new HashSet<>(Arrays.asList(INTER, DIFF, NONE));
+    private final static String STACK_RESOURCE = "AWS::CloudFormation::Stack";
+    private final static String STACK_SET_RESOURCE = "AWS::CloudFormation::StackSet";
+
 
     /**
      * Embedded Stack or StackSet is not allowed
      *
      * @param type Resource type
      */
-    private static void validateResource(final String type) {
+    private static void validateResource(final String type, boolean isSelfManaged) {
+
         switch (type) {
             case "AWS::CloudFormation::Stack":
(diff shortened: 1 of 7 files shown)


aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/util/Validator.java
```diff
@@ -24,18 +24,27 @@ public class Validator {
...
      * @param type Resource type
      */
-    private static void validateResource(final String type) {
+    private static void validateResource(final String type, boolean isSelfManaged) {
+
         switch (type) {
             case "AWS::CloudFormation::Stack":
+                if(!isSelfManaged){
+                    throw new CfnInvalidRequestException(
+                            String.format("Nested %s is not supported in service managed %s", STACK_RESOURCE,STACK_SET_RESOURCE));
```
aradven, line 42: Need to add the 'type' parameter back here. This format statement takes two parameters, you're only passing in one. Think this would appear something like:

"Nested AWS::CloudFormation::StackSet is not supported in service managed "

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/util/Validator.java
```diff
@@ -24,18 +24,27 @@ public class Validator {
...
+                }
+                break;
             case "AWS::CloudFormation::StackSet":
                 throw new CfnInvalidRequestException(
-                        String.format("Nested %s is not supported in AWS::CloudFormation::StackSet", type));
+                        String.format("Nested %s is not supported in %s", STACK_SET_RESOURCE,STACK_SET_RESOURCE));
```
aradven, line 47: Same as above

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/util/Validator.java
macdmerr: Thanks, It does have 2 params, I just forgot to add  a space between them.  should be fixed now.

aradven (approved):

jkim-jr (approved):

PR #108, opened 2023-04-03
[ModuleVersion] Return empty string instead of null for DocumentationUrl

MalikAtalla-AWS:
*Description of changes:*
I made a mistake in my last PR. I can't remove DocumentationUrl from the readOnlyProperties, because that would enable users to specify it in the template which is not how the attribute is intended to be used. The contract tests are complaining that we're not always returning the documentationUrl. Removing it would be backwards-incompatible, so to adhere to the resource type contract we'll return an empty string when the module doesn't have a documentationUrl.

### DIFF
--- a/aws-cloudformation-moduleversion/aws-cloudformation-moduleversion.json
+++ b/aws-cloudformation-moduleversion/aws-cloudformation-moduleversion.json
@@ -66,6 +66,7 @@
   "readOnlyProperties": [
     "/properties/Arn",
     "/properties/Description",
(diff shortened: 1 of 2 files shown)


lejiati (commented):
So `DocumentationUrl` will always be returned, but depends on input, it could be an empty string?

lejiati (approved):

jtting-amazon (approved):

kristian-d (approved):

PR #109, opened 2023-04-14
Pre- and Post-script for ModuleVersion Contract Tests V2

MalikAtalla-AWS:
*Description of changes:*
This adds the necessary setup logic for running the Contract Tests independently of any manual one-time setup steps.


aws-cloudformation-moduleversion/contract-tests-artifacts/pre-script/buildspec.yml
```diff
@@ -0,0 +1,20 @@
...
+  build:
+    commands:
+      - echo Build started on `date`
+      - aws cloudformation create-stack --stack-name moduleversion-ctv2-prescript --template-body file://prescript.yml
+      - aws cloudformation wait stack-create-complete --stack-name moduleversion-ctv2-prescript
+      - echo "Created stack moduleversion-ctv2-prescript"
+      - aws cloudformation list-exports | jq -r '.'
+      - export BUCKET_NAME=$(aws cloudformation list-exports |jq -r '.Exports[] | select(.Name | contains("ModuleVersionContractTestBucket")).Value')
+      - echo "${BUCKET_NAME}"
+      - zip -r contract-test-module.zip fragments/ schema.json
```
cdavidxu-hub, line 15: I encountered a similar issue with putting S3Objects within buckets for my contract tests as well that also resulted in the "Unrecognized Token" issue. It seems contract tests are very finicky when it comes to types of files and naming in the contract-test-artifacts folder and I think it will only accept .yml files. I ended up creating the file inline in the buildspec like below
`touch index.py && echo "print('hello world')" >> index.py`
`zip python.zip index.py`
`aws s3api put-object --bucket $bucketName --key python.zip --body python.zip`
Not sure if you would be able to try something similar to see if that helps.

aws-cloudformation-moduleversion/contract-tests-artifacts/pre-script/fragments/sample.json
```diff
@@ -0,0 +1,11 @@
+{
```
cdavidxu-hub, line 1: What is this template used for?

cdavidxu-hub (commented):
Left a suggestion for your buildspec

aws-cloudformation-moduleversion/contract-tests-artifacts/pre-script/buildspec.yml
MalikAtalla-AWS: This part was actually working fine for me. If I can, I'd like to avoid having to inline these two files. It would become very hard to read and we'd lose syntax highlighting and such.

aws-cloudformation-moduleversion/contract-tests-artifacts/pre-script/fragments/sample.json
jtting-amazon: believe this is the module fragment

MalikAtalla-AWS: This is a Module. The Contract tests will call the resource which will register this Module in the registry. The content of the module doesn't really matter as long as it's valid cloudformation, because we're not creating a stack with it. We're only testing the ModuleVersion resource. (It's all a bit meta, I know).

jtting-amazon (approved):

aws-cloudformation-moduleversion/contract-tests-artifacts/pre-script/buildspec.yml
cdavidxu-hub: Makes sense, I should have clarified - I was able to place the S3Object into the bucket using a file in my path as well - the prescript step was successful. However this always caused an issue for me that seems similar to the error you're facing(Unrecognized token). I know, they seem completely unrelated and I was confused as well, but the inline solution was what ended up working for me. Feel free to try it, I know its not ideal and theres no root cause so I understand the apprehension.

cdavidxu-hub (approved):

aws-cloudformation-moduleversion/contract-tests-artifacts/pre-script/buildspec.yml
MalikAtalla-AWS: Discussed offline. I don't see the connection, but will investigate.

PR #110, opened 2023-04-14
Disable ModuleVersion ListHandler

MalikAtalla-AWS:
The ListHandler of the ModuleVersion resource does not behave as expected. It always returns an empty response because the paging logic is not implemented properly. Fixing it is no trivial, because we need to iterate over an unknown number of lists (the list of modules and for each module the list of versions), each of which has an iteration token. So, we should disable it to reduce confusion for customers. It's better to return a clear message saying the list handler is not supported than returning an incorrect response. We'll work on fixing the handler in the meantime.

### DIFF
--- a/aws-cloudformation-moduleversion/aws-cloudformation-moduleversion.json
+++ b/aws-cloudformation-moduleversion/aws-cloudformation-moduleversion.json
@@ -100,12 +100,6 @@
         "cloudformation:DeregisterType",
         "cloudformation:DescribeType"
       ]
-    },
-    "list": {
-      "permissions": [
-        "cloudformation:ListTypes",
-        "cloudformation:ListTypeVersions"
-      ]
     }
   },
   "additionalProperties": false


aws-cloudformation-moduleversion/aws-cloudformation-moduleversion.json
```diff
@@ -100,12 +100,6 @@
         "cloudformation:DeregisterType",
         "cloudformation:DescribeType"
       ]
-    },
-    "list": {
```
jtting-amazon, line 104: Does disabling this cause any problems with the contract tests? If not, ill approve the PR

MalikAtalla-AWS: I've talked to the team owning the Contract Tests and they agreed to exclude tests for us that require list or update operations (we also don't have an Update handler)

cdavidxu-hub (approved):

jtting-amazon (approved):

PR #123, opened 2024-05-15
Add ConcurrencyMode Support for StackSet Resource

eperryAWS:
*Description of changes:* This PR adds the ConcurrencyMode parameter to the AWS::CloudFormation::StackSet resource.

### DIFF
--- a/aws-cloudformation-hookdefaultversion/pom.xml
+++ b/aws-cloudformation-hookdefaultversion/pom.xml
@@ -35,7 +35,7 @@
         <dependency>
             <groupId>software.amazon.cloudformation</groupId>
             <artifactId>aws-cloudformation-rpdk-java-plugin</artifactId>
-            <version>[2.0.0, 3.0.0)</version>
+            <version>[2.0.0, 2.1.0)</version>
         </dependency>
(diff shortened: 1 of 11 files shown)


aws-cloudformation-hookdefaultversion/pom.xml
```diff
@@ -35,7 +35,7 @@
         <dependency>
             <groupId>software.amazon.cloudformation</groupId>
             <artifactId>aws-cloudformation-rpdk-java-plugin</artifactId>
-            <version>[2.0.0, 3.0.0)</version>
+            <version>[2.0.0, 2.1.0)</version>
```
oshardik, line 38: is this version downgrade intended?

eperryAWS: These are intended. It's less of a downgrade and more to prevent an upgrade. The latest versions are incompatible with the JDK version we use, so I needed to restrict it to the latest that was compatible.

oshardik: gotcha, that sounds good

aradven (approved):

PR #126, opened 2025-02-24
Throw retryable exceptions if another operation is in progress when a…

aradven:
…ttempting to delete stack instances

*Issue #, if available:*

*Description of changes:*
A recent change to the RPDK changes the behavior of progress events such that they will no longer retry. We need to explicitly throw a retriable exception to maintain the current behavior


ksamyak-aws (approved):
Looks good to me.

oshardik (approved):
Looks good

PR #127, opened 2025-08-18
Allow out-of-band tags for stackset to remain during updates

ksamyak-aws:
*Issue #, if available:*

*Description of changes:*

Allows out of band tags to remain during update stack set when performed by the resource provider. This makes it so the tags on the stackset resource itself are not overridden.

### DIFF
--- a/aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/translator/RequestTranslator.java
+++ b/aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/translator/RequestTranslator.java
@@ -11,13 +11,18 @@ import software.amazon.awssdk.services.cloudformation.model.GetTemplateSummaryRe
 import software.amazon.awssdk.services.cloudformation.model.ListStackInstancesRequest;
 import software.amazon.awssdk.services.cloudformation.model.ListStackSetOperationResultsRequest;
 import software.amazon.awssdk.services.cloudformation.model.ListStackSetsRequest;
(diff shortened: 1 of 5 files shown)


patinti (approved):

aws-cloudformation-stackset/src/main/java/software/amazon/cloudformation/stackset/translator/RequestTranslator.java
```diff
@@ -106,7 +111,17 @@ public static DeleteStackInstancesRequest deleteStackInstancesRequest(
...
+            final Map<String, String> tags,
+            final List<Tag> currentTags) {
+        // Aggregate all current tags
+        Set<Tag> allTags = new HashSet<>(currentTags);
+        Set<Tag> currentTagSet = translateToSdkTags(tags);
+        allTags.addAll(currentTagSet);
+        // Remove the one's that were in the previousTags but not in tags
+        Set<Tag> previousTagSet = translateToSdkTags(previousTags);
+        previousTagSet.removeAll(currentTagSet);
+        allTags.removeAll(previousTagSet);
```
aradven: This is a little hard to follow, a lot of stuff being added/removed. Do we not just need to set the tags to targetTags + out of band tags? So something like:

outOfBandTags = currentTags - previousTags
tagsToSet = targetTags + outOfBandTags

aradven (changes requested):

aradven (approved):

patinti (approved):

